PR #7, opened 2019-12-04
Connectors homepage

runnabro:
Changes:
* Add Connectors page
* Fix select focus transition
* Remove ToC sidebar when ToC is not present (and not a doc)

Todo:
- [x] Handle popovers
- [x] Browser testing
- [x] Connector article page
- [x] Add remaining Connector images

Screenshots:
Screenshot 2019-12-05 16 04 50
Screenshot 2019-12-05 16 03 28
Screenshot 2019-12-05 16 03 11


runnabro:
@mojavelinux I'm ready for ya. Everything should be stubbed out in place with all the interaction bits. 

Non-functional bits:
**Homepage:**
* Clicking on table row to go to connector page
* Dynamic table content
* Dynamic menu content
* Search and filters

**Connector Article:**
* Version selector
* Dynamic tier popover
* Back to connector homepage link
* Dynamic avatar + h1 for articles
* Dynamic categories, runtime, links, tier

src/partials/connectors-popular.hbs
```diff
@@ -0,0 +1,134 @@
+<section class="connectors-popular">
+  <header class="header-connectors-title">Popular Connectors</header>
+  <div class="flex popular-cards">
+    <a class="card" href="//docs.mulesoft.com/connectors/salesforce/salesforce-connector">
+      <img
+        alt="Salesforce"
+        class="card-img"
+        height="50"
+        src="/_/img/logos/salesforce.png"
```
mojavelinux: @runnabro When making a reference to an image from a template, please use `{{uiRootPath}}` instead of `/_`.

src/partials/connectors-tablerow.hbs
```diff
@@ -0,0 +1,51 @@
+<tr class="tr" tabindex="0">
+  <td class="td">
+    <img class="img" src="/_/img/icons/connectors/icon-amazon-s3.svg" width="30" alt="Amazon S3 logo">
```
mojavelinux: @runnabro Do we really want a logo per connector, or should we use the vendor logo?

src/partials/connectors-menu.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="js-connector-menu" style="display: none;">
```
mojavelinux: The connectors-menu div will need to be inside the row or it will need to be populated from information in the row when it appears. Right now, it's static, so it's the same content for every row.

src/partials/connectors-tablerow.hbs
runnabro: That's my personal preference—but technically the logo is what appears in Exchange. I'll raise the issue with the team and see what they think.

sleslie23:
@runnabro FYI, after talking to @mojavelinux yesterday, we need to base this PR off of the `sandbox` UI branch. Currently, this branch is in synch with the `master` branch. This will enable us to test the connector landing page changes in our sandbox environment.

sleslie23:
@mojavelinux FYI, I got the following message when I attempted to change the base branch:

"Some commits from the old base branch may be removed from the timeline, and old review comments may become outdated."

Just wanted to double check with you before I change it.

mojavelinux:
I think it's fine. It's likely just a warning. If it does screw something up, Tony should be able to force push the branch and it will restore the state.

mojavelinux:
@sleslie23 @runnabro I have done a proof of concept for the connectors landing page that retrieves data dynamically from the UI model. However, the fact that the connectors landing page is taking over the default layout is a blocker. That needs to be resolved before I can submit a PR to the PR. The solution is to make a new layout named connectors-home.hbs (or similar) and leave default.hbs unmodified.

src/partials/connectors-tablerow.hbs
sleslie23: @runnabro @mojavelinux I would suggest using the vendor logo also. Not sure why we have different logos in Exchange. I'll follow up with Meera (the connectors PM).

sleslie23:
@mojavelinux FYI, the sandbox UI is up and building in the sandbox env.

src/partials/connectors-tablerow.hbs
sleslie23: Actually, ignore my previous comment. Meeta just mentioned that we cannot use some of the vendors icons because of licensing issues. The Exchange team has dealt with this issue in the past. As long as we use the icon that is in Exchange, we should be ok.

sleslie23:
@mojavelinux @runnabro  Hey Dan! I added a new layout for the connectors landing page. I don't see any changes to the default.hbs layout in this PR, however. Do we also need to add logic to partials/article.hbs and partials/main.hbs to use this layout? AFAIK that's what we do for the 404 page.

mojavelinux:
@sleslie23 The change is in the partial that is used by the default layout, so effectively it modifies the default layout (just not directly).

src/partials/connectors-tablerow.hbs
mojavelinux: To be clear, I'm not concerned with whether its a logo or icon or png or svg. What I'm looking for is a consistent pattern so that we can compute the path to it from the metadata. I'm also curious whether the logo is that of the vendor (which would need to be defined in the metadata) or that of the connector (which can be based on the component name).

sleslie23:
> @sleslie23 The change is in the partial that is used by the default layout, so effectively it modifies the default layout (just not directly).

Got it. Looks like that change would require some rearchitecting of the structure of the connectors partials. It probably make sense to have @runnabro do that. FYI, he mentioned earlier that he won't be able to get to it until next week.

mojavelinux:
You've got it.

@runnabro If you can confirm you won't me modifying the PR this week, I can go ahead and shift the partials around so they don't take over the default layout. Then I can proceed with my proposal to make the pages read data from the UI model.

sleslie23:
> You've got it.
> 
> @runnabro If you can confirm you won't me modifying the PR this week, I can go ahead and shift the partials around so they don't take over the default layout. Then I can proceed with my proposal to make the pages read data from the UI model.

Hey @mojavelinux I don't think Tony will be able to get to it until Monday. If you can shift the partials around this week that would help us avoid delays.

src/partials/connectors-tablerow.hbs
runnabro: I think Steven is saying that they prefer the connector logo.

runnabro:
> It would be helpful if we stuck with either SVG or PNG for the icons / logos

All the logos will be png for consistency (since not all companies provide SVGs for their logos), but all icons will be SVG (we should be able to depend on Connector icons being SVGs since they come from Exchange). Is that okay?

mojavelinux:
Great! @runnabro, I'll apply my changes then send a PR to your PR. That will give you a chance to see what changes are incoming. Then you can accept the changes using the merge button (or request changes).

mojavelinux:
Once https://github.com/mulesoft/docs-site-ui/pull/13 is merged, I think this PR should be merged this into the sandbox branch so we can get a bundle URL to start testing.

PR #10, opened 2019-12-19
Make connectors homepage dynamic

mojavelinux:
* create dedicated layout for connectors home
* repurpose default layout to auto-detect connector article
* use display version in nav and page meta if available (required to support Mule 3 connector version)
* update build to discover partials in subdirectories
* remove directory segment when calling partial in subdirectory
* configure quicklinks menu to work per row; only trigger on click
* make templates for connectors landing dynamic
* add required data to UI model

What's still missing from this update is a sample connector article page and the required template logic to make the connector version sidebar dynamic. I'll do that in a separate PR.


mojavelinux:
This PR targets the connectors-homepage branch (so merging it will add commits to #7.

fermujica:
Thanks @mojavelinux !
Should @runnabro review this before merging or are we OK with adding this to his `connectors-homepage` branch ?

mojavelinux:
I just wanted @runnabro to see what was coming before modifying his PR. If he gives the :+1:, then it's ready to merge.

mojavelinux:
@fermujica @runnabro I have implemented the search filter on the connectors homepage. I also fixed some problems in the template for the popular connectors.

I'm going to defer work on the connectors article page and sidebar until a separate PR since that was scoped out in the project plan as a separate task. The focus of this PR is the connectors homepage.

@sleslie23 @fermujica How is the release notes link determined for a connector? That's the one piece that is missing.

Aside from that, the connectors homepage is fully functional. It will require some updates to the component descriptors for the connectors, but I'll address that when I update the pipeline for the sandbox environment.

src/partials/connectors/connectors-popular.hbs
```diff
@@ -1,134 +1,43 @@
...
-        alt="SAP"
-        class="card-img"
-        height="50"
-        src="{{uiRootPath}}/img/logos/sap.png"
-        width="98"
-      >
+    {{/with}}
+    {{else}}
+    <a class="card" href="#">
+      <img src="{{@root.uiRootPath}}/img/icons/connectors/icon-{{connectorName}}.svg" alt="{{connectorName}} Logo" class="card-img" width="71" height="50">
```
runnabro: For the popular connectors, I was thinking the path should be:
`src="{{@root.uiRootPath}}/img/logos/{{attributes.page-vendor-name}}.png"`

runnabro: Also (I'll do this in my PR), I think we'll have to remove setting the width here because the logos will have different widths (but keep the same height).

src/js/07-connectors.js
```diff
@@ -117,5 +54,39 @@
...
+      onShow: function (instance) {
+        instance.popper.classList.remove('hide')
+      },
+      onShown: function (instance) {
+        instance.popper.classList.add('shown')
+      },
+    })
+  }
+
+  function toArray (collection) {
```
runnabro: This is not currently used and gave me an err when building

src/partials/connectors/connectors-popular.hbs
mojavelinux: That's fine with me. I just wanted to present you with the various options.

src/js/07-connectors.js
mojavelinux: Yep, no longer needed.

PR #24, opened 2020-02-05
Fix various problems in navigation code

mojavelinux:
* only attempt to relativize a root-relative URL
* fix scope of local variables in toggleNav function
* don't fail if navigation for product is missing
* use display version if defined for current version


mojavelinux:
@sleslie23 These are fixes in preparation for the connectors navigation. It doesn't add any new functionality, so you should be able to merge it right away.

PR #25, opened 2020-02-06
DOCSINFRA-161 Connectors navigation

mojavelinux:
This PR implements the navigation for connectors. That entails moving the navigation for the individual connectors to the bottom of the Connectors navigation children. These items still behave like top-level navigation items, but only appear when the Connectors item has been activated.


mojavelinux:
The PR depends on #24 and needs to be rebased once that PR is merged.

mojavelinux:
@sleslie23 This PR is now ready for review. You should be able to try it with a local playbook that builds the sandbox site.

PR #30, opened 2020-02-25
more fixes for connectors

mojavelinux:
- fix bug in filter when connector only has a Mule 3 version
- use display version to detect Mule 3 connector
- drop mule- prefix from runtime version
- update preview model


mojavelinux:
@sleslie23 These updates a required to build the latest output of the connectors migration.

PR #34, opened 2020-03-21
Track ToC usage via Amplitude

runnabro:
Sends Amplitude event for ToC link click and touch events with props for doc title, url, and url hash.


sanderson-sfdc:
Not doing; consensus is that it won't work as written.

PR #38, opened 2020-05-07
Update header from marketing to include Connect banner

fermujica:


src/css/globals/vars.css
```diff
@@ -64,7 +64,7 @@
   --z-gdpr: 20;
   --z-modal-header: 16;
   --z-modal: 15;
-  --z-nav-mobile: 14;
+  --z-nav-mobile: 1000;
```
sanderson-sfdc, line 67: Why did this have to bump up to override the other ones?  Did you test if search or whatever is a modal still works with this change?

src/css/specific/header.css
```diff
@@ -1,13 +1,33 @@
 /* overrides for marketing header */
-.ms-com-content-header {
+body.flex .ms-com-content.ms-com-content-header {
   border-bottom: 1px solid var(--aluminum-3);
   box-shadow: 0 3px 15px rgba(0,0,0,.025);
+  z-index: 1000;
```
sanderson-sfdc, line 5: Rather than hard-coding this, can you use a variable?

src/css/specific/navigation.css
```diff
@@ -22,7 +22,7 @@
     position: sticky;
     top: 0;
     transform: translateX(0);
-    z-index: var(--z-nav);
+    z-index: 0;
```
sanderson-sfdc, line 25: As above, can you use a variable here?

src/css/globals/vars.css
fermujica: As of today, we can't test the search bar unless we build this on stgx.

sanderson-sfdc: Ah, right.  Do what you think is best, then.

fermujica:
Turned the z-index values into variables. 

The values were suggested by marketing team while we troubleshooted some overlapping issues with these changes.

PR #41, opened 2020-05-22
DOCSINFRA-256 add meta tag.

sleslie23:
Adding metatag to "verify" Google Search Console.


mojavelinux:
See my other comment. I recommend defining this as a site key, which you would access as `site.keys.googleSiteVerification`.

mojavelinux:
It's also better to use `#with` instead of `#if`. The `#with` block combines a property reference and truthy conditional.

```
{{#with site.keys.googleSiteVerification}}
<meta name="google-site-verification" content="{{this}}">
{{/with}}
```

(Notice I also used an unclosed tag, which is preferred in HTML5)

We've also been avoiding blank lines above and below a statement in the template, so I recommend removing those for consistency.

cheungaryk:
@sanderson-sfdc is this still needed? I see that the change Dan suggested in https://github.com/mulesoft/docs-site-ui/pull/41#issuecomment-632909355 is already in the codebase https://github.com/mulesoft/docs-site-ui/blob/9de663a32ad5b0096bed39fa4eed6770bbe2c34d/src/partials/head-meta.hbs#L7

PR #44, opened 2020-06-11
DOCSINFRA-253 Merge sandbox UI into production

mojavelinux:
* enhance preview build and align with Antora 2.3
* general style changes
* update header and header styles
* add templates, CSS, and JavaScript to support listing page and individual pages for connectors


mojavelinux:
One thing we should do before merging is updating the dependencies. It's always a good idea to do this when making a major update so the code is as current as possible.

PR #52, opened 2020-07-16
DOCSINFRA-306 Allowing local edit links

sanderson-sfdc:
DOCSINFRA-306 Allowing local edit links


mojavelinux:
There's only one problem. The conditionals need to be reversed. That's because editUrl is always set, but fileUri is only set if the file comes from a local git worktree.

See https://gitlab.com/antora/antora-ui-default/-/blob/master/src/partials/toolbar.hbs

sanderson-sfdc:
@mojavelinux - like this?

sanderson-sfdc:
And, hopefully last question, Jenkins builds the new package, I don't have to, right?

mojavelinux:
That is correct. By merging into the master branch, the CI build will generate a new release. Then, to make use it, you need to update the release number in the playbook file for the site.

PR #55, opened 2020-09-03
docsinfra279 - updating UI bundle for Japanese to support new connectors

sanderson-sfdc:


sanderson-sfdc:
Screen Shot 2020-09-03 at 5 41 26 PM
Screen Shot 2020-09-03 at 5 48 38 PM

Ran the link checker before/after and it has the same number of broken links, so nothing new due to the ui change.

sanderson-sfdc:
I also requested the L10N team to re-translate the English header-content and footer-content partials.  I'll add those later.

mojavelinux:
If it looks good to you, :+1: from me. Let me know if you need help getting something working.

sleslie23 (approved):

PR #59, opened 2020-09-24
DOCSINFRA-364: Allow partial builds for connectors in the beta branch

sanderson-sfdc:


sleslie23 (approved):
LGTM, however I'm no JavaScript expert.

PR #60, opened 2020-11-02
Update marketing header and footer to remove newsletter sign up option

fermujica:
Upate the marketing footer and header to remove the **Sign up for our newsletter** sign, which is no longer an available link.

this


sanderson-sfdc (approved):
LGTM!

PR #67, opened 2021-02-26
npm audit fix fixed 99 vulnerabilities and updated package-lock.json

agansop:


sanderson-sfdc (approved):
Verified against latest docs-site-playbook build.

PR #72, opened 2021-03-29
Update marketing header to fix misaligned "Contact" button.

fermujica:
This PR fixes the misaligned "Contact". "Free trial", and "Login" options in the site's header:

.Before:
current

.After:
new


sanderson-sfdc (commented):
Looks like there are a bunch of text changes as well; was that intentional?

For example, Product becomes Product, the description for the Anypoint Platform has changed, etc.

fermujica:
Yes. All those updates come from marketing so that we have the same header as mulesoft.com and developer.mulesoft.com (minus the search option next to "Login").

sanderson-sfdc (approved):

PR #80, opened 2021-07-08
DOCSINFRA-527 Updating HTML generation to make result valid with CSP

sanderson-sfdc:
Removed _almost all_ of the inline styles (Antora bug filed for remaining ones)
Added a nonce for the inline javascript ( https://scotthelme.co.uk/csp-nonce-support-in-nginx/ ) 
Upgraded the package.json files


agansop (approved):
LGTM

agansop (dismissed):
LGTM

PR #90, opened 2021-12-01
DOCSINFRA-655: Implement Survey banner 

fermujica:
[DOCSINFRA-655 - KB/DOCS Implement banner in MuleSoft Documentation site](https://www.mulesoft.org/jira/browse/DOCSINFRA-655)

## Description

This PR adds the survey banner to the documentation site UI bundle.

Screen Shot 2021-11-26 at 13 35 29

## Testing

This change was tested in local environments using Browserstack. 

Browsers: Latest Chrome, Firefox, Safari, and Edge versions.
Platforms: macOS, Windows 10 & 11, iOS, Android

## Pending Items

* We must add the proper link to the survey in the "Take the survey!" button. ✅ 
* We must make the banner responsive as it displays incorrectly in all mobile platforms. ✅


sanderson-sfdc:
@fermujica based on the pending items, it doesn't seem like a good idea to merge this.  Do you want help accomplishing those goals?

fermujica:
The banner is now responsive, and I added the link to the survey. 

@sanderson-sfdc Should we hide the banner if the site is not being rendered in stgx or prod? I vote yes on that. Otherwise, the survey banner would be rendering on local builds as well and could interfere with the metrics if someone clicks it outside the docs site. 

Unfortunately, to fully test this, I need to release a new version of the UI, and test it in a Jenkins build.

kevintroller (approved):

sanderson-sfdc:
@fermujica - building a new UI bundle isn't dangerous, so go ahead and do that.  

Regarding lower environments versus prod, with the upcoming change to the environment (we're doing a rolling release of the new infrastructure starting tomorrow), I'm not sure it's worth the time.

PR #99, opened 2022-02-14
[W-10592577] add article footer

cheungaryk:
ref: [W-10592577](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000laXxJYAU/view)

### Bugfixes

- Fix error `site.js:4 Uncaught TypeError: Cannot read properties of null (reading 'querySelector')` on connector pages by adding article-footer, which contains the missing/null elements

### Enhancements

- in readme, update references of node 10 to node 12 which is the current LTS version we are using


sanderson-sfdc (commented):
Does this resolve the issues with a topic that uses the https://github.com/mulesoft/docs-site-ui/blob/master/src/layouts/connectors-home.hbs layout?

cheungaryk:
> Does this resolve the issues with a topic that uses the https://github.com/mulesoft/docs-site-ui/blob/master/src/layouts/connectors-home.hbs layout?

Yes. I also added the article-footer section to connectors-home.hbs partial which should work now. In my local testing, the footer shows on both the connectors home and individual connector pages.

sanderson-sfdc (approved):
Nice job; I built the bundle locally and tested both with preview and running a build against the built bundle and the error is gone.

PR #100, opened 2022-02-21
W 10728368 - survey banner

cheungaryk:
ref: [W-10728368](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000qTEcjYAG/view)

### Enhancements

- added survey to all connector pages (moved the survey section into article.hbs)
- added `margin:auto` to survey CSS to align it to the center
- I have noticed that in mid screen size, the survey banner doesn't look right, usually it is too long for the section. Therefore, I standardized the CSS for the survey banner for sm/md/lg screen sizes to match what's there for the sm screen size

---

### How to test

- run `gulp preview` and open localhost:5252/. Scroll down to see that survey banner is aligned to the center of the main viewport section. Drag the left/right edges of the browser to decrease width and see that the banner will change to a square-like shape
- go to the [Salesforce connector](http://localhost:5252/connector.html) and see that the survey button is present
- click on the "take the survey" button opens the userzoom page in a new tab
- http://localhost:5252/connectors.html is the only page that does not have the survey. Unsure if it needs to be there, please clarify


fermujica:
Thank you for working on this.

I have three observations about this banner:

- I agree that we should not have the survey banner displayed on the Connectors landing page.
We shouldn't display it on the 404 page either: 
Screen Shot 2022-02-22 at 13 15 01
(Note that the location of the banner conflicts with the message).

- The banner is too close to the "Was this article helpful?" survey that displays only on production. I think that we don't have this survey in the new site's infrastructure. @sanderson-sfdc, should we consider fixing this or are we deprecating the "was this article helpful?" survey?
Screen Shot 2022-02-22 at 13 22 04

- The "Take the survey" button text is not fully centered:
 Screen Shot 2022-02-22 at 13 26 32
Kevin and I were looking at this, and found out that removing the `button-text` class and adding the text as part of the `a` element, centered the text properly (see [dff3eb6](https://github.com/mulesoft/docs-site-ui/commit/dff3eb6db9d1691211d2bb03a31bd91edeee4010))
 Screen Shot 2022-02-22 at 13 17 58

src/partials/article.hbs
```diff
@@ -16,3 +16,4 @@ If you typed the URL of this page manually, please double check that you entered
 {{/if}}
 </article>
 {{> article-footer}}
+{{> survey}}
```
fermujica, line 19: Adding the survey to the article-footer partial instead of the article partial fixes the banner being too close to the "was this article helpful?" banner, and also removes it from the 404 page.
https://github.com/mulesoft/docs-site-ui/commit/bcb5e865f1d5a2d4386069a0b1aee635d4ad91b9

But does not remove it from the connectors landing page. 
Question: Is the article-footer partial needed in the connectors landing page?

cheungaryk: Nice find. 

Yes, the article-footer partial is needed to resolve an error described in #99

cheungaryk: I got a solution to exclude the footers from 404 pages: https://github.com/mulesoft/docs-site-ui/pull/100/commits/b74b86ce899b898a504dd2ba4b8c20502335e4cd

Also, updated CSS to adjust the spacings and I think it looks decent: https://github.com/mulesoft/docs-site-ui/pull/100/commits/cbb75b8ea67890f1d47376cf6b9bc7d8613d39ac

kevintroller (approved):

sanderson-sfdc (approved):
Great collaboration.  Who's going to take the task to update the playbook once the bundle is generated?

cheungaryk:
> Great collaboration. Who's going to take the task to update the playbook once the bundle is generated?

I can do it (once I figured out how from the docs)

cheungaryk:
@sanderson-sfdc or @kevintroller can you give this another 👍🏼 ?

src/partials/footer-content.hbs
```diff
@@ -73,7 +73,7 @@
         </section>
         <section class="footer-bottom">
             <div class="footer-inside">
-                <p class="footer-copyright"><span>&copy;2021</span> MuleSoft LLC, a Salesforce company</p>
+                <p class="footer-copyright"><span>&copy;1</span> MuleSoft LLC, a Salesforce company</p>
```
sanderson-sfdc, line 76: How does `&copy;1` turn into the 2022 copyright?

sanderson-sfdc (changes requested):

src/partials/footer-content.hbs
cheungaryk:
```suggestion
                <p class="footer-copyright"><span>&copy;2022</span> MuleSoft LLC, a Salesforce company</p>
```

sanderson-sfdc (approved):

PR #101, opened 2022-02-23
[W-10592498] add searchbox focus

cheungaryk:
ref: [W-10592498](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000laXUcYAM/view)

## Fixes

Per ESLint, `{key}.keyCode` is deprecated. Using the newer `{key}.key` instead.

## Enhancements

When the left navigation lane's search box is clicked to show the search modal, move the mouse to focus on the modal's search box. (Note that I am not able to test this one locally)


sanderson-sfdc (approved):
Thanks; it is frustrating that we can't try it locally, but it looks right to me.

PR #116, opened 2022-03-11
W-10736467 Fix survey position for Connectors pages

kevintroller:
Fixed survey position for Connectors pages

Currently live incorrect position:
Screen Shot 2022-03-11 at 11 45 13 AM

New fixed position:
Screen Shot 2022-03-11 at 11 45 32 AM

Could you double check this on your end to make sure it looks good? The localhost preview shows only Salesforce Connector and http://localhost:5252/salesforce-connector-authentication.html page was the only page that has an available TOC on the right to test the correct position.
Can this be tested with a different connector page? How is that done?

### DIFF
--- a/src/layouts/default.hbs
+++ b/src/layouts/default.hbs
@@ -30,6 +30,7 @@
       <div class="js-toc">
         <nav class="toc-menu"></nav>
       </div>
+{{> survey-toc}}
     </aside>
   </div>
   {{else}}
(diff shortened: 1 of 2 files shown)


cheungaryk:
Instead of adding survey-toc to default.hbs, can you move 

```html
<div class="js-toc">
  <nav class="toc-menu"></nav>
</div>
```

into connectors-sidebar.hbs?

the reason for the move is that for non-connector pages, the toc <div> block is already included in body.hbs, so it doesn't need to stay in default.hbs. It should be placed in the connectors-sidebar.hbs because it's part of that side bar.

I have validated this locally that making the above change will not remove the TOC or header from any non-connector pages.

cheungaryk (approved):

PR #122, opened 2022-03-23
[W-10592498] fix focus on modal

cheungaryk:
ref: W-10592498

## Bug fixes

- fix the logic which allows the search box focus to work: 
  - the previous logic was not working because the searchBox did not exist when it was initialized, so it was null when the `focus()` function was called, and it failed. The element was also not the right one
  - the current logic adds an interval check to make sure the right `input` element exists, and then focus on it
- revert `e.key` to `e.keyCode`. It broke the esc function

## Validation

1. grab the Coveo token here: https://github.com/mulesoft/docs-site-playbook/blob/master/Jenkinsfile#L7 and export it into an environment variable via `export COVEO_ACCESS_TOKEN=<token>`
2. run `gulp preview` and open http://localhost:5252 in a web browser
3. click the search box on the upper-left corner (it shows up because a valid Coveo token exists). Your focus should jump to the textbox of the modal
4. tap the esc button on your keyboard. The modal is gone


sanderson-sfdc (approved):
LGTM, but I'm not sure how to test it.

PR #125, opened 2022-03-30
fix overflow menus in connectors table so they don't show by default

mojavelinux:


mojavelinux:
@sanderson-sfdc This seems to have broken when the change was made to respect the CSP policy. But it's a super simple fix.

In the future, there is a simpler way to hide elements by default. You can use the hidden attribute instead of a hidden class. See https://caniuse.com/hidden

sanderson-sfdc (approved):

cheungaryk:
LGTM, but this PR can't be merged because it has diverged too much from the base branch? 

@sanderson-sfdc opened a separate PR #127 which has the same content and is easily mergeable, so let's merge that one instead (I also approved it). I will close this PR.

mojavelinux:
As I mentioned in the other thread, if you need me to rebase an MR, just ask. I will do it as quickly as I can.

PR #129, opened 2022-04-28
[W-11071581] adjust z-nav

cheungaryk:
ref: W-11071581

adjust z-nav so it goes behind the marketing banner that sometimes appears. Tested manually with Safari incognito window (more detailed steps in GUS ticket)


agansop (approved):
approved

PR #130, opened 2022-04-29
[W-11071630] adjust survey TOC CSS for Safari

cheungaryk:
ref: [W-11071630](https://gus.my.salesforce.com/a07EE00000vxGusYAE)

## Bug Fix

- remove `type="button"` from survey TOC element. It was breaking the element appearance in Safari

## Enhancement

- slight adjustments on the survey element CSS so it looks better on Safari

## Validations

These screenshots are taken from Safari. The layouts are not perfect but we can fix them later.

Screen Shot 2022-04-29 at 2 15 54 PM

Screen Shot 2022-04-29 at 2 16 00 PM

### DIFF
--- a/src/css/components/survey.css
+++ b/src/css/components/survey.css
@@ -23,7 +23,7 @@ div.survey-icon {
 }
 
 div.survey-title {
-  display: block;
+  display: inline-block;
   font-family: var(--sans-serif);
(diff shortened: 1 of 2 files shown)


agansop (approved):
Approved

PR #134, opened 2022-05-06
W-11108816 Change mulesoft.com to www.mulesoft.com to meet CSP requir…

sanderson-sfdc:
…ements


agansop (approved):
Approved

sanderson-sfdc:
The only change is the reference to the salesforce-logo.svg; changing the host to www.mulesoft.com

PR #135, opened 2022-05-06
[W-11108816] remove helmet

cheungaryk:
ref: W-11108816

Per Web Team, remove the helmet banner for now since it has some CSS issues. Web Team is expected to fix the CSS issues and ask us to put the banner up again.

Note that in local testing (`gulp preview`), you will see a gap where the helmet used to be in mobile view. This is acceptable, because the new CSS (which will deploy in an hour) will adjust and remove the gap.


sanderson-sfdc (approved):
I'll trust the change without a local build; please test on stgx (on Jenkins, select the "build" parameter)

PR #152, opened 2022-07-05
[W-10950274] added deprecated banner

cheungaryk:
ref: W-10950274

Added deprecated banner for versions that are marked deprecated. To use it, add this to the antora.yml of a branch of a docs repo:

```yml
asciidoc:
  attributes:
    deprecated: true
```

Please see https://www.loom.com/share/66084827c09942228dc24ce4c4a9f657 for detail.


IsaacEldridge (approved):
@cheungaryk Thanks for this! A question:
In the video, you said there is no "x" (to close the banner). This by design? I think it's the right move.

cheungaryk:
@IsaacEldridge yes, we should test this in a beta site. I want to validate how the beta site looks on mobile...

cheungaryk:
> @cheungaryk Thanks for this! A question: In the video, you said there is no "x" (to close the banner). This by design? I think it's the right move.
> 
> Before we roll this out to the team, can we try it on the RTF docs? I can try it out tomorrow.

I saw that in your mockup (https://mulesoft.slack.com/archives/G02H6FZJD/p1647981441247239), there is an "x" on the lower right corner. If it is not needed, I won't add it. 😛 

I will test this on a beta site tomorrow!

PR #156, opened 2022-07-12
[W-10950274] deprecated container v2

cheungaryk:
ref: W-10950274

This is an alternative version of the deprecated banner (now is actually a "container"). It is at the top of each page of a deprecated version, and the position is not fixed. Otherwise, similar color and feel.

See https://www.loom.com/share/a4cff85a28fd417084dc2523c5357003 for the look and feel.


sanderson-sfdc:
Thanks for doing this; I assume that Isaac and/or you are going to be using this to share around with stakeholders?  If so, can you make this a draft PR?

IsaacEldridge (commented):
Thanks for putting this together so quickly. Does the text have to be hardcoded, or could the writer go in and change the text depending on their use case? Ideally, we're only going to allow a few approved messages to go in the box. 

Can we make it left-aligned? 
 I'm going to share this PR with the team. Thanks!

agansop (dismissed):

cheungaryk:
- add the second canned message listed in https://docs.google.com/document/d/1LTi12VKfHoZUBQleApO4QPo6_dUYep7jREIwT1CJOEg/edit
- update the attribute names to `olderVersion` and `noLongerUpdated`
- fix the "latest version" links:
  - they now work for beta sites
  - if a page in an older version doesn't have an equivalent page in the latest version, it will redirect to the latest version's index page instead

See https://www.loom.com/share/a4cff85a28fd417084dc2523c5357003 for detail. @IsaacEldridge to review

src/partials/notice-banner.hbs
```diff
@@ -0,0 +1,23 @@
...
+  <div class="notice-banner">
+    <p>
+      {{#if page.componentVersion.asciidoc.attributes.noLongerUpdated}}
+      We are no longer updating this version of the documentation.
+      {{else}}
+      {{#if page.componentVersion.asciidoc.attributes.olderVersion}}
+      You’re viewing an older version of the documentation.
+      {{/if}}
+    {{/if}}
+      Switch to the
```
IsaacEldridge, line 16:
```suggestion
      You can switch to the
```

src/css/specific/notice-banner.css
```diff
@@ -0,0 +1,35 @@
+.notice-banner {
```
IsaacEldridge, line 1: Can you make sure it's left-aligned?

IsaacEldridge (approved):

src/css/specific/notice-banner.css
cheungaryk: it is left-aligned, please see attached video to validate.

I have removed all the images of the earlier designs (which are center-aligned) to avoid confusion. The video is now the source of truth

cheungaryk:
> @cheungaryk - is there a way to validate this with `bundle preview`?

I have not found a way for preview to recognize the attributes in ui-model.yml. I only tested using a local build with test-3 UI bundle

IsaacEldridge (approved):

PR #169, opened 2022-07-29
New nav with groups

cheungaryk:


cheungaryk:
see https://github.com/mulesoft/docs-site-ui/pull/170

PR #173, opened 2022-08-04
[W-11549560] spread single nav item

cheungaryk:
ref: [W-11549560](https://gus.lightning.force.com/a07EE000013HtpRYAS)

add new playbook variable, **spreadSingleItem**, to make it more flexible on how the nav should behave when there is only a single nav item inside a nav group. The default behavior is false, but if set to true inside the playbook, the subpages in the single nav item will be promoted to nav items themselves (this is how the "Getting Started" section is structured).

if spreadSingleItem is not set or set to `false`:

Screen Shot 2022-08-04 at 11 18 33 AM

if spreadSingleItem is set to `true`:

Screen Shot 2022-08-04 at 11 18 46 AM

https://www.loom.com/share/cf3d98a0602a45299e21657dab0cdd76


sanderson-sfdc (approved):
I thought a bit about the default behavior and I agree, defaulting it to false is the right choice for the use case.

cheungaryk:
> I thought a bit about the default behavior and I agree, defaulting it to false is the right choice for the use case.

Yup, forgot to clarify that doing it this way means less work for the writers because they don't need to specify the variable at all.

PR #178, opened 2022-08-10
[W-11267884] dataweave tryme button

cheungaryk:
ref: W-11267884

- add a "try me" button to dataweave code snippets. The button will link to Dataweave playground (blank), or a pre-populated playground if there is a referenced Github file
- renamed some of the class elements and files


sanderson-sfdc (approved):
Brilliant; I noticed the mathjax example in "gulp preview" isn't working for me, but that might just be a local issue.

PR #192, opened 2022-09-21
[W-11668954] refresh Japanese site header and footer

cheungaryk:
ref: [W-11668954](https://gus.lightning.force.com/a07EE000015GmwlYAC), https://github.com/mulesoft/docs-site-playbook/pull/993

- merge the latest UI changes in the English site into the Japanese UI bundle. There are too many changes to list, but here are a few:
  - breadcrumbs on screen sizes
  - new antora nav tree reorganization
- update header and footer to match closely to the current ones on https://www.mulesoft.com/jp
  - I kept the old language selector at the upper right corner because I couldn't figure out how to make the new language selector start at defaulting at Japanese. And it isn't really a selector anyway, more like a link to go back to the English site
  - footer language selector is not in scope - there is another ticket for it, and actually the new footer doesn't have a language selector anymore as it is moved to the upper right corner of the header
 
Screen Shot 2022-09-21 at 4 55 49 PM

Screen Shot 2022-09-21 at 4 38 35 PM

Reviewers, please review the following at the minimum:
- language selector
- the left nav headers like "Getting Started" and "Docs by Product". They are in English - should they be translated?


agansop (dismissed):

sanderson-sfdc:
When I preview, I see English in the header, Japanese in the footer, and no language picker. 

Screen Shot 2022-09-22 at 9 15 12 AM
Screen Shot 2022-09-22 at 9 15 18 AM

cheungaryk:
> When I preview, I see English in the header, Japanese in the footer, and no language picker.
> 
> Screen Shot 2022-09-22 at 9 15 12 AM Screen Shot 2022-09-22 at 9 15 18 AM

Nice catch. I had to remove the automatic header content update: [c1570fa](https://github.com/mulesoft/docs-site-ui/pull/192/commits/c1570fa99f364202766262efde58c562c0a79bac)

sanderson-sfdc:
It appears that the attached is the language picker; I need to be able to see it in my language, not Japanese, to use it.

Screen Shot 2022-09-22 at 11 45 42 AM

I just realized we don't do that on our English site either, so I'm going to approve this, but we need a bug to make this more functional on both sites.

sanderson-sfdc (approved):

cheungaryk:
> It appears that the attached is the language picker; I need to be able to see it in my language, not Japanese, to use it.
> 
> Screen Shot 2022-09-22 at 11 45 42 AM
> 
> I just realized we don't do that on our English site either, so I'm going to approve this, but we need a bug to make this more functional on both sites.

I don't understand what you mean here, but will wait to see more context in the bug ticket.

PR #193, opened 2022-09-22
[W-11799442] remove version selector labels from archive site

cheungaryk:
ref: W-11799442

in archive site, in the version selectors, remove the labels "current version" and "previous versions" since they don't make sense there.

Screen Shot 2022-09-22 at 3 52 43 PM

To test, here are the recommended steps:
- open docs-site-ui, switch to this branch, run `gulp bundle` to create a local UI bundle
- open docs-site-playbook, switch to the archive branch, create a local playbook, change ui.bundle.url to the local UI bundle, run `make gen-site playbook=<playbook path>` to build it, look at the built page and make sure that there are no version labels in the version selectors, and no other UI defects
- still on docs-site-playbook, switch to the default branch, create a local playbook, change ui.bundle.url to the local UI bundle and remove most of the repos, run `make gen-site playbook=<playbook path>` to build it, look at the built page and make sure that the version labels **are present** in the version selectors, and no other UI defects


src/js/01-nav.js
```diff
@@ -534,5 +538,16 @@
...
+    const t = navData.groups
+      ? navData.groups.filter((a) => hasArchiveInName(a.homeTitle) || hasArchiveInName(a.title))
+      : []
+    return t.length > 0
+  }
+
+  function hasArchiveInName (name) {
+    return String(name).toLowerCase().includes('archive')
+  }
+
```
cheungaryk, line 551: at the moment, I don't have a non-flaky way to tell whether it is archive site or prod site. I come up with the method to check that the group headers have 'archive' listed in their names. Not the best way, but it will do for now...

sanderson-sfdc: Isn't `site-profile`  available at that time?  See https://github.com/mulesoft/docs-site-ui/blob/b5de752e22c6e8302d5ae30c90d314919a6d3819/src/partials/banner.hbs#L10

cheungaryk: I don't think that variable is available for the javascript. Let me do a research and see if there is a better way to do this

cheungaryk: I have simplified the method using `window.location.host`: [44dc7ab](https://github.com/mulesoft/docs-site-ui/pull/193/commits/44dc7ab8c702128612c9b1f03fa13c3fd439c49c). Much more reliable since it checks whether the base URL (archive.docs.mulesoft.com) contains the word 'archive' or not.

sanderson-sfdc (approved):

PR #194, opened 2022-09-23
[W-10592902] Hide search suggestions at small screen sizes to make filter visible

cheungaryk:
ref: W-10592902

Note: this has some dependencies to the work at https://github.com/mulesoft/docs-site-ui/pull/191, specifically the adjusted length of the search box which no longer blocks the filter in medium and large screen sizes.

- add logic so that the auto-suggestions are hidden when "Search Docs" is clicked initially, so that the suggestions don't block the filters (with 1000 max tries to prevent infinite loop if elements are not found)
- updated header content (this step was automatic)

https://www.loom.com/share/eefc0ae3204c478191691858fc04fd3e


agansop:
@cheungaryk This PR has a conflict, do you still want it to be reviewed ?

cheungaryk:
> @cheungaryk This PR has a conflict, do you still want it to be reviewed ?

@agansop the conflict was caused by merging #191, in the marketing header. I have resolved it 😄

agansop (approved):

PR #203, opened 2022-10-11
[W-10592536] add current version indicator + notice banner close button

cheungaryk:
ref: W-10592536

- add a blue dot indicator for the current version in the left nav version selector. The indicator has a tooltip that tells users what it means
- upgrade a library called tippy.js to the latest version
- add close button ("x") to deprecated notice banners to allow users to close it temporarily. It will display again when page is refreshed, no cookie is involved

current version indcator:
https://www.loom.com/share/5a84587a82384ae79d23a282f70fb259

notice banner close button:
https://www.loom.com/share/34e6a6a7a1934f5f86aaad183a6279f1

### DIFF
--- a/src/js/01-nav.js
+++ b/src/js/01-nav.js
@@ -246,20 +246,15 @@
 
   function createNavVersionDropdown (navItem, componentData, page) {
     var versions = Object.values(componentData.versions)
-    var currentVersionData =
-      versions.length > 1
-        ? versions.find(function (version) {
-          return !version.prerelease
-        }) || versions[0]
-        : versions[0]
+    var currentVersionData = getCurrentVersionData(versions)
     var navVersionDropdown = createElement('.nav-version-dropdown')
     navVersionDropdown.addEventListener('click', trapEvent)
     var navVersionButton = createElement('button.button.nav-version-button')
     var activeVersion = componentData.name === page.component ? page.version : currentVersionData.version
     var activeDisplayVersion = componentData.versions[activeVersion].displayVersion
-    navVersionButton.appendChild(
-      createElement('span.nav-version', { dataset: { version: activeVersion } }, activeDisplayVersion)
-    )
+    var navVersion = createElement('span.nav-version', { dataset: { version: activeVersion } }, activeDisplayVersion)
+    if (activeVersion === currentVersionData.version) addCurrentVersionIndicator(navVersionButton)
+    navVersionButton.appendChild(navVersion)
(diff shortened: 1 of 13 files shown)


src/js/01-nav.js
```diff
@@ -304,17 +302,63 @@
...
+  }
+
+  function createCurrentVersionIndicator (tabIndex) {
+    const currentVersionIndicatorSpan = document.createElement('span')
+    currentVersionIndicatorSpan.setAttribute('role', 'tool-tip')
+    currentVersionIndicatorSpan.classList.add('tooltip-dot')
+    currentVersionIndicatorSpan.setAttribute('tabindex', tabIndex)
+    tippy(currentVersionIndicatorSpan, {
+      arrow: tippy.roundArrow,
+      content: 'This is the latest version.',
```
sanderson-sfdc, line 340: This should be in a labels file so we can get this translated into Japanese.

src/partials/notice-banner.hbs
```diff
@@ -9,6 +9,7 @@
       page.componentVersion.asciidoc.attributes.olderVersion))
     }}
     <div class="notice-banner">
+      <button class="notice-banner-close-button" title="Close the notice banner">&times;</button>
```
sanderson-sfdc, line 12: I know I'm being picky on this, but we should move this to a labels file, too.  

I'm sure we have other strings that aren't properly translated, but we should do this properly anyway.

src/partials/notice-banner.hbs
```diff
@@ -9,6 +9,7 @@
...
       {{#if page.componentVersion.asciidoc.attributes.endOfLife}}
       <p>This version of the product has reached
```
sanderson-sfdc, line 14: I know you didn't change this, but since it's close to the new work, maybe you could move this into a labels file, too.

sanderson-sfdc (changes requested):

cheungaryk:
@sanderson-sfdc I know that we have an initiative to merge the Eng and JP UI bundles (which is actually like 70% done in #195 but in a blocked state for 2 more weeks), but it feels out of scope for this ticket and was never communicated as a requirement prior.

My response is to create a new user story for separating out the languages in a label file (is it the same as a map?), and then probably add the work to #195.

sanderson-sfdc:
That's the right action, @cheungaryk , thanks for doing that.  I should have created the story instead of adding a requirement here.

PR #212, opened 2022-11-10
[W-12030048] make right sidebar sticky at the bottom of the page

cheungaryk:
ref: W-12030048

Screenshot 2022-11-10 at 9 27 47 AM


sanderson-sfdc (approved):
Nice!  And such an easy fix, too.

PR #300, opened 2023-02-10
[W-12522437] new way to set EOL/support status banner

cheungaryk:
ref: W-12522437

### New way to set the text of the ~~eol~~ support status banner

previously, to set the support status banner, we add an attribute in a doc repo's branch's **antora.yml** file as described in https://confluence.internal.salesforce.com/pages/viewpage.action?pageId=507972975#UsingBanners-DocVersionsThatNoLongerReceiveUpdates. Example:

```yml
asciidoc:
  attributes:
    olderVersion: true
```

but as we get more and more statuses (some are not even related to EOL, so we renamed it to status), the attribute naming convention isn't scalable. So we have come up with a new way to do this:

```yml
asciidoc:
  attributes:
    supportStatus: old|noLongerUpdated|extendedSupport|eolSupport|eolScheduled|eolReached
```

A different status generates a different set of wording. If the status is invalid, the banner will not show at all. Here are some examples that are invalid:

``` yml
supportStatus: sdafjkdaslfsaklf # does not exist
supportStatus: extendedsupport # wrong case (should be capital S in the middle)
supportStatus: eolSuport # misspelled (missing a p in the middle)
```

To maintain stability, the old attributes have been retained until we have updated all occurrences (for example, there are a few in docs-run-fabric), then we will deprecate the old attributes.

The [custom banner message](https://confluence.internal.salesforce.com/pages/viewpage.action?pageId=507972975#UsingBanners-UseCustomTextBanners) (via page attribute `page-notice-banner-message`) will remain available and will not be deprecated.

### new handlebar helpers

Added 4 helpers: `switch`, `case`, `default`, and `if`. 

- switch/case/default are all part of the switch-case function in traditional programming, and these are used by the supportStatus use case above
- the new `if` helper allows the `else if` syntax which helps making the handlebar templates cleaner - see d2aa3626f275ef80e2fbf9d10e081731b489f1fe as an example. For further information, see [handlebar documentation on conditionals](https://handlebarsjs.com/guide/block-helpers.html#conditionals).

### other changes

- update marketing content to the latest (manually removing skip links that came with the headers because they conflicted with our own skip links)
- rename some of the banner files


src/partials/support-status-banner/banner-for-eol-scheduled.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="notice-banner">
```
cheungaryk, line 1: @hannanelson you can look at the wording of all the banner-for-* files starting from this one.

src/partials/support-status-banner/banner-for-eol-scheduled.hbs
```diff
@@ -0,0 +1,18 @@
...
+    <button class="notice-banner-close-button" title="Close notice banner">&times;</button>
+    <p>
+        {{#if page.latest.url}}
+        This version of the product
+        {{else}}
+        This product
+        {{/if}}
+        is scheduled for
+        <a class="notice-banner-link" href="https://www.mulesoft.com/legal/versioning-back-support-policy"
+            target="_blank" rel="noopener noreferrer">End of Life.</a>
```
cheungaryk, line 11: @hannanelson you can look at the wording of all the banner-for-* files starting from this one.

hannanelson (approved):
LGTM

PR #318, opened 2023-02-24
[W-12601706] update banners

cheungaryk:
ref: W-12601706

@hannanelson see new banner attributes in https://github.com/mulesoft/docs-site-ui/blob/11e889bc60acb7d5f9c4b815c7744f1b45140270/src/partials/support-status-banner/support-status-banner.hbs. Note that I updated the attribute value `eolReached` to `eol` to simplify it.


hannanelson (dismissed):
LGTM

src/partials/support-status-banner/support-status-banner.hbs
```diff
@@ -8,9 +8,17 @@
 {{#case 'old'}} {{> banner-for-old}} {{/case}}
```
hannanelson, line 8: can we change this to `oldVersion`?

cheungaryk: yes. Updated

hannanelson (approved):

PR #342, opened 2023-03-13
[W-12653092] added required files for Asciidoctor Tabs

agansop:


cheungaryk:
sorry for the delayed response. The reason why I have not approved the PR is because I am not sure that's supposed to look like. In my mind, I assume it should look something like this:

image

I don't see any images in the original repo that shows what it should look like either. Maybe you can ask Tasha what she is looking for exactly, and if your screenshot is not what she wants, then come back to troubleshoot this.

agansop:
@cheungaryk Tasha is checking with the SA of the project and will get back to me

cheungaryk:
see #352

PR #353, opened 2023-03-23
[W-12690315] adjusted maintenance banner design

agansop:


src/css/specific/top-banner.css
```diff
@@ -22,16 +22,14 @@
 
   & img {
     filter: grayscale(100%) brightness(0);
-    height: 17px;
-    vertical-align: middle;
+    height: 22px;
+    margin: auto 10px auto 10px;
```
cheungaryk, line 26: can be shortened as `margin: auto 10px`

src/partials/top-banner.hbs
```diff
@@ -3,10 +3,10 @@
 page.componentVersion.asciidoc.attributes.announcementMessage)
 }}
 <div class="top-banner flex {{#if page.componentVersion.asciidoc.attributes.announcementMessage}}announcement{{/if}}">
-  <div class="container wrapper">
+  <div class="container wrapper flex">
     {{#if page.componentVersion.asciidoc.attributes.announcementMessage}}
+     <img src="{{@root.uiRootPath}}/img/icons/important.svg" alt="">
```
cheungaryk, line 8: nice change. You also need to do the same for the beta site banner (line 13-14 of this file. Basically just switch those lines to get  out of <p>)

src/css/globals/grid.css
```diff
@@ -17,6 +17,7 @@
 
 .flex {
   display: flex;
+  justify-content: center;
```
cheungaryk, line 20: if you add this property here, it will apply the property to all elements with the flex class, which is not desirable.

In line 50 of this file, there is another class named `justify-center` which adds the same property to any elements with that class. 

Therefore, you should take this property out of `.flex`, and then add the `justify-center` class to the .container.wrapper div in top-banner.hbs. That should give the `justify-content: center` property to the div without affecting `.flex`

cheungaryk (changes requested):

cheungaryk (approved):

src/css/specific/top-banner.css
agansop: fixed

src/partials/top-banner.hbs
agansop: fixed

src/css/globals/grid.css
agansop: fixed

PR #363, opened 2023-04-04
[W-12653092] tabs look & feel adjustments

cheungaryk:
ref: W-12653092

implemented everything that was described in https://www.loom.com/share/c19623eab7e44d2cb6e76cfc7d837e11.

- instead of importing the default JS and CSS files, copied those files and pasted in our UI bundle instead, this allows us to customize them
- force tabindex for all tabs to be 0, thus allowing all the tabs to be reachable by keyboard
- adjusted color for the active tab to be more visible
- handled error which caused the external link script to break if an element doesn't have a parent element

TODO:
- adjust the appearance of the various states (focus, hover, active). Waiting for design mockups to complete


cheungaryk:
new video showcasing how tabs work (including the various states): https://www.loom.com/share/0185e91ea8cb4c00b9bcd548f55f7dac

agansop (approved):

PR #525, opened 2023-09-14
[W-14073070] Added logic to switch content from en to jp

agansop:
current version and previous version are updated when the site in is Japanese.
Screenshot 2023-09-14 at 1 58 34 PM

Collapse content is translated to Japanese
Screenshot 2023-09-14 at 2 01 59 PM

Expand content is translated to Japanese
Screenshot 2023-09-14 at 2 01 42 PM


cheungaryk (changes requested):

src/js/01-nav.js
```diff
@@ -8,9 +8,13 @@
   const languageMap = {
     en: {
       title: 'Home',
+      current_ver: 'Current version',
+      previous_ver: 'Previous versions',
```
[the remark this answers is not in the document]
agansop: @cheungaryk  What about putting currentVersion and previousVersion I think it makes more sense.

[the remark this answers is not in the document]
cheungaryk: it would make sense if the words are 'Previous version' instead of 'Previous version**s**'. With 1 being single and 1 being plural, they are quite different...

src/partials/toc.hbs
```diff
@@ -4,9 +4,8 @@
 </div>
 {{> github}}
 {{> survey iconName="arrow-down"}}
-<h1 class="toc-title">On this page:</h1>
-<div class="aside-toc">
-  <nav class="toc-menu" aria-label="Skip to page content sections"></nav>
+<h1 class="toc-title">{{#if (eq (site-profile) 'jp' )}}このページの内容{{else}}On this page{{/if}}:</h1><div class="aside-toc">
```
cheungaryk, line 7: please move `<div class="aside-toc">` to its own line. Other than that LGTM

cheungaryk (changes requested):

src/partials/toc.hbs
cheungaryk: @agansop I saw that you removed that line altogether. It is needed, just need to be put in a new line to make it easier to read.

agansop: @cheungaryk should be fixed now

cheungaryk (approved):

cheungaryk (approved):

cheungaryk (approved):

bjoson-mulesoft (approved):

PR #537, opened 2023-09-21
[W-14110499] added code-builder

agansop:


src/helpers/release-notes.js
```diff
@@ -1,5 +1,5 @@
 const maxNumofItems = 10
-const components = ['composer', 'release-notes']
+const components = ['composer', 'release-notes', 'docs-code-builder']
```
cheungaryk, line 2: This should be the component name, not the repo name.

The component name can be found in the antora.yml file in the repo. https://github.com/mulesoft/docs-code-builder/blob/37638a41d8ca3be9200dc0b1cc6f7c3a1572c59a/antora.yml#L1

cheungaryk (changes requested):

src/helpers/release-notes.js
agansop: Updated

cheungaryk (approved):
Approved, but for practice, I would suggest testing it locally using the video I sent you on Slack.

PR #543, opened 2023-09-26
[W-14188182] updated ch1 to ch

agansop:
Screenshot 2023-09-26 at 2 35 56 PM


cheungaryk (changes requested):
thanks for the screenshot. The screenshot is showing that some buttons are taller than the others and they shouldn't be. 

Upon inspection, it's because of the new buttons have top and bottom margins:

Screenshot 2023-09-26 at 11 51 34 AM

If you remove those margins from the buttons, then all the buttons would look normal again. If you want to double make sure that the height doesn't change, you can set the height property as well. All existing buttons right now show the height of 32px.

cheungaryk:
> thanks for the screenshot. The screenshot is showing that some buttons are taller than the others and they shouldn't be.
> 
> Upon inspection, it's because of the new buttons have top and bottom margins:
> 
> Screenshot 2023-09-26 at 11 51 34 AM
> 
> If you remove those margins from the buttons, then all the buttons would look normal again. If you want to double make sure that the height doesn't change, you can set the height property as well. All existing buttons right now show the height of 32px.

to clarify on `remove those margins`: one way to do that is to set margin to 0.

agansop:
> > thanks for the screenshot. The screenshot is showing that some buttons are taller than the others and they shouldn't be.
> > Upon inspection, it's because of the new buttons have top and bottom margins:
> > Screenshot 2023-09-26 at 11 51 34 AM
> > If you remove those margins from the buttons, then all the buttons would look normal again. If you want to double make sure that the height doesn't change, you can set the height property as well. All existing buttons right now show the height of 32px.
> 
> to clarify on `remove those margins`: one way to do that is to set margin to 0.

Set margin to 0

Screenshot 2023-09-26 at 3 38 46 PM

cheungaryk (approved):

PR #561, opened 2023-10-18
[W-13315036] added js script for review site

agansop:


src/partials/head/head-scripts.hbs
```diff
@@ -87,5 +87,17 @@ window.settings_timer=setTimeout(function () {_vwo_code.finish() },settings_tole
...
+    "openSidebar": true,
+    "openLoginForm": true
+  }
+</script>
+
+<script type="application/json" class="js-hypothesis-config">
+  {
+    "sidebarAppUrl": "https://dev-docs-review-h.kdev.msap.io/app.html"
+  }
+</script>
```
cheungaryk, line 101: why have 2 different script blocks? Why not just:

``` javascript
<script type="application/json" class="js-hypothesis-config">
  {
    "openSidebar": true,
    "openLoginForm": true,
    "sidebarAppUrl": "https://dev-docs-review-h.kdev.msap.io/app.html"
  }
</script>
```

agansop: It was provided by Mehdi I haven’t tested with only one block

agansop: Let me test first with one block.

agansop: Testing with one block does not work, h apps doesn’t open at all

cheungaryk (approved):

PR #568, opened 2023-10-23
[W-14080808] Fix: Code snippets display incorrectly in callout lists

jdomingu:
Code snippets in callout lists should display the border and light background that all other inline code snippets display.

Note that an alternate fix would have been to change the CSS selector to `& code` directly (to match all code snippets in the doc). I confirmed that works as well, but I was wary of unintended consequences, so I kept things scoped it down. If the broader scope is preferred, LMK and I can do that instead.

Note that I went ahead and confirmed during testing that code snippets already display correctly (with the background) in all other lists and tables.

**Before fix**

Screenshot 2023-09-06 at 15 15 30

**After fix**

Screenshot 2023-10-23 at 12 17 59 PM

### DIFF
--- a/src/css/adoc/doc.css
+++ b/src/css/adoc/doc.css
@@ -305,7 +305,7 @@
   }
 
   & p code,
-  .calloutlist code {
+  & .calloutlist code {
     background: var(--aluminum-1);
     border: 1px solid var(--aluminum-3);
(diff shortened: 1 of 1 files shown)


cheungaryk (approved):
validated via local build. This scope is fine

PR #576, opened 2023-10-26
[W-14314888] Centralize UI strings (first pass)

jdomingu:
This change does the following:

* Moves a subset of our current UI strings to a central location in `src/locales/messages.json`. (Note that moving all the strings we currently have, including those in templates, seemed like a much bigger work item, so this is just a first pass.)
* Inserts the messages into our JS bundle at buildtime with the `gulp-replace` plugin.
* Provides a way to get loc messages from JS.
* Provides a way to get loc messages from HTML.

I've also put together some internal documentation for all this on our wiki at [Localization for Docs Site UI Strings](https://confluence.internal.salesforce.com/x/Vj0pL).

**Testing**

* Verified that we get the locale properly from the `html lang` attribute.
* Verified that every UI string moved to the messages file displays properly in `en` and `jp` locales. I've confirmed this in the preview site and in a full build of our `jp` playbook.
* Verified that when a message is missing in the current locale, we fall back to the message in the English locale.
* Verified we throw an error properly when attempting to get UI messages that do not exist in the current locale or the fallback locale.

Here's a couple of examples:

Screenshot 2023-10-24 at 3 57 44 PM

Screenshot 2023-10-26 at 10 48 26 AM


cheungaryk (approved):
very cool! Having a centralized file for all the translated strings makes it all very manageable, and the [wiki page](https://confluence.internal.salesforce.com/x/Vj0pL) makes it look very easy to implement.

Validated via local build.

PR #581, opened 2023-10-30
[W-14086638] add version to breadcrumbs + separate Trending Topics with Coveo UA/Search 

cheungaryk:
ref: W-14086638

- add version to breadcrumbs so that it's easier for users to see what the version they are currently in
  Screenshot 2023-10-30 at 11 31 26 AM
- separate the Coveo scripts into 2 blocks: 1 for Trending Topics, and another for Coveo User Analytics+Search. This fixes the issue where internal beta sites cannot displaying trending topics because they weren't built with the Coveo API search key (on purpose, to avoid hitting usage quota)


jdomingu (approved):
Nice! That's very helpful.

PR #588, opened 2023-11-03
[W-14113620] set grid=none

agansop:
Screenshot 2023-11-07 at 11 22 04 AM


src/css/adoc/table.css
```diff
@@ -208,6 +208,20 @@ div.panels table {
     overflow: auto;
     width: 100%;
 
+    &.grid-none {
+      & > * > tr {
+        border: none !important;
+
+        & > td {
+          border: none !important;
```
cheungaryk, line 216: don't need `!important` here.  Ok to keep `!important` in line 213 because it overrides the line 377 (it's technically possible to not use important there, but it may get too complicated)

src/css/adoc/table.css
```diff
@@ -208,6 +208,20 @@ div.panels table {
...
+        }
+      }
+    }
+
+    &.frame-shown {
```
cheungaryk, line 221: is it frame-shown or frame-show?

cheungaryk (changes requested):
Please see https://salesforce-internal.slack.com/archives/C0458B71FFF/p1698954810510779?thread_ts=1694474667.803329&cid=C0458B71FFF for the requirements. 

Implemented:
- frame-show (or frame-shown?)
- grid-none

Not yet implemented:
- grid-rows
- grid-cols

Finally, please add some more tables in table.adoc so that it is easier to validate.

src/css/adoc/table.css
agansop: Looks like she doesn't want the inner rows and columns so I am not sure why do we have to implement grid-rows and grid-cols as it is overriding grid-none. grid-none should be sufficient

src/css/adoc/table.css
agansop: Updated to frame-show

src/css/adoc/table.css
```diff
@@ -204,9 +204,42 @@ div.panels table {
...
+    &.grid-cols {
+      & tr {
+        border-top: unset !important;
+      }
+      
+      & td {
+        border-left: 1px solid #cacbcc;
+        border-right: 1px solid #cacbcc;
+      }
+    }
```
cheungaryk, line 238: This does not work because it causes outer borders.

To remove the outer border, do the following:
- change `td` to `td+td` to specify for all <td>s except for the first (leftmost column)
-  remove border-right property

then for clean up, use `var(--aluminum-4)` instead of #cacbcc.

cheungaryk (changes requested):
Running the requirements from https://salesforce-internal.slack.com/archives/C0458B71FFF/p1698954810510779?thread_ts=1694474667.803329&cid=C0458B71FFF:

1. not specifying the frame - no frames ✅ 
2. `frame=show` - frames are showing, but different color than the usual borders 🟡 
3. `grid=rows`: looks like example 4 ✅ 
4. `grid=cols`: not 100% look like example 5 - has outer borders ❌ 
5. `grid=none`: no grids ✅

src/css/adoc/table.css
```diff
@@ -204,10 +204,41 @@ div.panels table {
...
+      }
+      
+      & td+td {
+        border-left: 1px solid var(--aluminum-4);
+      }
+    }
+
+    &.frame-show {
+      border: 1px solid var(--aluminum-4);
+    }
```
cheungaryk, line 240: consider sorting these alphabetically to make it easier to find. In this case, frame-show show be higher than the grid-* ones

src/css/adoc/table.css
```diff
@@ -204,10 +204,41 @@ div.panels table {
     border-top: unset;
     display: block;
     font-size: 14px;
-    margin-top: var(--lg);
+    margin-top: var(--lg) 0 0;
```
cheungaryk, line 207: margin-top -> margin

cheungaryk (changes requested):
minor changes. Also this comment was never addressed: https://github.com/mulesoft/docs-site-ui/pull/588#discussion_r1385220718

cheungaryk (commented):
Don't want to keep repeating this...but this comment still has not been addressed. https://github.com/mulesoft/docs-site-ui/pull/588#discussion_r1385220718

Is the comment not clear? It's about making it easier to validate your changes by creating 1 table for each of the requirements. So for example, 1 table for grid=rows, 1 table for grid=cols, **and then clearly label them visually** in some way. Preferably similar to how the other tables are labeled.

preview-site-src/elements/table.adoc
```diff
@@ -405,4 +405,94 @@ Here is a second paragraph in the same cell.
...
+|Cell in column 1, row 4
+|Cell in column 2, row 4
+|Cell in column 3, row 4
+
+|Cell in column 1, row 5
+|Cell in column 2, row 5
+|Cell in column 3, row 5
+|===
+
+=== combination of frame and grid
```
cheungaryk, line 483: maybe specify this as frame-show and grid-none to make it more clear

cheungaryk (commented):
last minor comment...otherwise looks good to me. All requirements have been satisfied and clearly demonstrated in the preview site.

preview-site-src/elements/table.adoc
agansop: @cheungaryk updated!

cheungaryk (approved):
LGTM!

PR #591, opened 2023-11-06
[W-14093770] Apply One Salesforce Styles

jdomingu:
This change applies some of the base Salesforce styles, but does not touch or header and footer (as that work is on hold pending internal discussions).

Here are the changes to our styles:

- Add and import Salesforce fonts.
- Add new Salesforce colors.
- Update heading sizes, weights, colors, etc.
- Update other typography styles, including styles for paragraphs, links, etc.
- Update home page to remove dot patterns, use Salesforce colors, and button styles.

Note: This change does not touch the feedback form styles as we received those directly from design.

Tested locally across screen resolutions.


preview-site-src/elements/miscellaneous.adoc
```diff
@@ -37,3 +37,12 @@ This is the same in-page cross reference:
 This is an in-page cross reference for the #bold ID: <<bold,Click me>>.
 
 xref:table.adoc[Go to the table page]
+
+== Headings
+
+= H1 Test Heading
+== H2 Test Heading
```
cheungaryk, line 44: I know this is for testing, but every time I run `npm run preview`, this triggers the error message "asciidoctor: ERROR: <stdin>: line 43: level 0 sections can only be used when doctype is book", which sounds scary even though it isn't. Maybe just refer it to the actual heading of this page, which happens to be an H1?

H2 can refer to the H2 headings too, so H3...H6 Test Headings can be nested under `== Headings`.

src/css/globals/typography.css
```diff
@@ -1,37 +1,77 @@
...
 
 h2 {
-  font-size: 32px;
+  font-size: 40px;
+  line-height: 48px;
 }
 
 h3 {
-  font-size: 28px;
+  font-size: 32;
```
cheungaryk, line 58: 32px

cheungaryk (changes requested):
looks straight forward to me. 1 place needs to be fixed + 1 minor suggestion.

src/css/globals/typography.css
jdomingu: Thank you! Excellent catch.

preview-site-src/elements/miscellaneous.adoc
jdomingu: Absolutely fair. Removed the H1 to eliminate the error.

cheungaryk (approved):

PR #592, opened 2023-11-06
[W 13579027] Add aria-expanded for details elements

jdomingu:
This PR improves the accessibility of the details element by adding the `aria-expanded` attribute.

Note that I've renamed the `12-tip-svg-no-alt-text.js` file to `12-accessibility-fixes.js`, but it doesn't show very cleanly in the diff unfortunately.


src/js/12-accessibility-fixes.js
```diff
@@ -0,0 +1,34 @@
...
+    setDetailsAriaExpanded(elem)
+  }
+
+  const setDetailsAriaExpanded = (elem) => {
+    if (elem.hasAttribute('open')) {
+      elem.setAttribute('aria-expanded', true)
+      return
+    }
+    elem.setAttribute('aria-expanded', false)
+  }
```
cheungaryk, line 22: this can be rewritten as:

```
const setDetailsAriaExpanded = (elem) => {
  elem.setAttribute('aria-expanded', elem.hasAttribute('open'));
}
```

cheungaryk (changes requested):

src/js/12-accessibility-fixes.js
jdomingu: Well I feel silly now. Yes, that is a great call.

src/js/12-accessibility-fixes.js
```diff
@@ -14,11 +14,7 @@
   }
 
   const setDetailsAriaExpanded = (elem) => {
-    if (elem.hasAttribute('open')) {
-      elem.setAttribute('aria-expanded', true)
-      return
-    }
-    elem.setAttribute('aria-expanded', false)
+    elem.setAttribute('aria-expanded', elem.hasAttribute('open'));
```
cheungaryk, line 17: this has an extra `;` which failed the CI/CD linting stage

cheungaryk (changes requested):

cheungaryk (approved):

PR #594, opened 2023-11-07
[W-14093770] Additional One Salesforce styling changes

jdomingu:
After the last PR #591 went out to the beta site, I did some additional testing with Clint and discovered a few areas for improvement.

Adding the following:

* Match the Trending topics link styles to the global link styles. (Missed this initially as I didn't have a Coveo key saved locally--took care of that this time.)
* Fix bug with footer size on homepage. (Removing the background dots had unintended consequences.)
* Update button styles (not just for landing page this time). I was initially cautious of affecting the feedback button styles, but after noticing that there's several topics like [this one](https://docs.mulesoft.com/general/api-led-design) that include button elements, I went ahead and updated the default button primary styles.
* Fix missing hover states.

### DIFF
--- a/src/css/components/button.css
+++ b/src/css/components/button.css
@@ -8,25 +8,17 @@
 
 .button {
   background: var(--tertiary);
-  border: 1px solid var(--aluminum-4);
-  color: var(--steel-2);
+  border: 1px solid var(--sds-g-color-blue-50);
+  color: var(--sds-g-color-blue-50);
   height: var(--nav-item-height);
   padding: var(--xs);
 
-  &:hover:not(:disabled) {
-    background-color: var(--light-blue);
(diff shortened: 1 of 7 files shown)


cheungaryk (commented):
Tested with local build and preview site. A few observations on the landing page:
- "More Resources" section
  - the dots are gone? I think it's intentional though give your PR description
  - hover state not obvious for the links
- Trending Topics section links - no hover statex

jdomingu:
Thanks, so much for being thorough! Added back in the missing hover states. The missing dots are indeed intentional--not sure why the guidance says to remove them, but it's part of the story. 

Please let me know if you find anything else! These are all great catches.

cheungaryk (approved):

PR #596, opened 2023-11-07
[W-11646839] add dataweave code syntax highlighting

cheungaryk:
ref: W-11646839

Adding dataweave code syntax highlighting! This helps dataweave customers by enhancing dataweave code readability.

Screenshot 2023-11-07 at 11 58 18 AM

In order to make this happen, I switched to [prismjs](https://prismjs.com/) to replace highlight.js. My implementation is based on Datastax's doc site (https://github.com/search?q=repo%3Ariptano%2Fantora-ui-docs%20prism&type=code). 

Validated via preview site and local build that:
- code syntax highlighting works for multiple languages
- the copy/paste button on the upper right corner works
- look & feel is _almost_ the same as what we currently have (some colors are slightly different)
- dataweave playground button on the upper right corner works


jdomingu (approved):
Awesome work! Local build looks great.

My only comment is concerning the `prism.js` file being committed directly into the repo. It looks like there's an npm package for it--if possible it would be nice to use that instead so that we can get automated dependency updates for it with our regular process. 

I clicked approve. If there's a good reason to include the file directly, then feel free to merge in as-is.

cheungaryk:
> Awesome work! Local build looks great.
> 
> My only comment is concerning the `prism.js` file being committed directly into the repo. It looks like there's an npm package for it--if possible it would be nice to use that instead so that we can get automated dependency updates for it with our regular process.
> 
> I clicked approve. If there's a good reason to include the file directly, then feel free to merge in as-is.

good point! I copied the exact implementation from Datastax which also used Antora, and I couldn't get prism working like highlight.js does (Antora officially supports highlight.js but [mentions that we can use other highlighters if needed](https://docs.antora.org/antora/latest/asciidoc/source/#source-highlighter:~:text=You%20can%20use%20other%20client%2Dside%20(also%20known%20as%20browser)%20libraries%2C%20such%20as%20prettify%20or%20prism.%20However%2C%20to%20use%20them%2C%20you%E2%80%99ll%20need%20to%20modify%20the%20UI%20to%20bundle%20and%20add%20the%20library%20to%20each%20page%2C%20like%20the%20reference%20UI%20does%20for%20highlight.js.)). I will merge for now...

PR #644, opened 2024-01-22
[W-14653224] make all external links open in new window

cheungaryk:
ref: W-14653224

The ask in the GUS ticket is to force all external links to open in new window, so that the writers don't have to add the `^` to each link.

There exists an script that adds an anchor icon to external links. Therefore, it is easy to just modify that script to capture all the links in the `.doc` and `.nav` areas (aka main viewport and left TOC), and check whether it is an external link (if the link does not start with `window.location.origin`). If it is an external link, then force it to open in a new window AND THEN add the anchor icon.

To validate, in the local preview site, I removed all the carets from the "Antora docs reference" links in multiple pages, and was able to validate that they still open in new windows + has the anchor icon.


jdomingu (approved):
🚀

PR #648, opened 2024-01-25
[W-12306459] adding JP preview site

cheungaryk:
ref: [W-12306459](https://gus.lightning.force.com/a07EE00001IjK1gYAF)

Adding a JP preview site. The content of this site is located in the **preview-site-src-jp/** directory, and the site is built by the command `npm run preview-jp`.

The reason for having a new directory is because in the original **preview-site-src/** directory, the key `siteProfile` in the file _ui-model.yml_ is a global variable that controls the language of the entire site. One can change the key's value to jp, but it is error-prone (for example, someone may change the value and check the new value in by accident), and merely changing that value doesn't give a full JP site experience because there are still a lot of English content. By having a new directory, we can add actual Japanese content adoc files there and lock the `siteProfile` value to jp.

An alternative solution that was tested was to add the Japanese content adoc files to **preview-site-src/**, and then add a page attribute to those pages, and finally update the partials to add logic to check said page attribute alongside checking whether the `site-profile` value is jp. The idea was dropped because 1) the additional logic makes the code harder to read and 2) it's error-prone, since the site-profile line becomes very long


jdomingu (approved):

cheungaryk:
@jdomingu please re-approve. I localized one of the words ("latest") and fixed some errors.

agansop (approved):

PR #652, opened 2024-01-31
[W-14585067] Replace icons with new Lume icons

jdomingu:
This PR replaces our current docs site icons with the Lume icons from the [Lume Design System](https://github.com/mulesoft/lume-design-system/tree/master/packages/lume-icons). I initially tried using the Lume-recommended approach to inserting icons as [custom elements](https://lume.mulesoft.com/?path=/docs/introduction-using-icon--docs), but we use icons in so many different ways that this proved more effort than it was worth.

Here are the ways we currently use icons:

* Creating icons from JS using the paths defined in `icondefs.js`--primarily used for left nav icons.
* Embedding icons with `img src="*.svg"`.
* With `<use xlink:href="#icon-part">` in some hbs partials.
* With the svg markup embedded directly in html in some hbs partials. 

As a result, making this change required the following work:
* Converting the svg icons from Lume to a JSON format we expect in `icondefs.js`.
* Replacing icons in `src/img/icons`.
* Replacing hardcoded svg in partials.

All of these required some massaging as:

* The sizes of icons didn't always match. I've inserted widths and heights where necessary and adjusted some of our css.
* The `use` functionality requires IDs in a handful of files which I've inserted manually.
* The icons embedded as images cannot use css variables (as the image is considered a separate document by the browser). As a result, I've hardcoded the color values in the svg files. 
* There wasn't always a 1:1 match for icons, so I've used my best judgement in selecting approximations. In particular the EU Control Plane and Runtime Fabric nav items did not have anything similar. 

As a nice to have, I also went ahead and selected icons for our nav items which previously did not have icons and were using the default Muley icon. (This includes the Contribute, Mule Gateway, Mule Runtime, Mule SDK, and MUnit nav items.) 

**Testing**

* Tested all the ways of inserting icons listed above.
* Ran full builds of docs-site-playbook and performed manual verification of various pages.
* Checked the display of icons at multiple resolutions.


cheungaryk (approved):
LGTM! Validated via local preview site.

PR #653, opened 2024-02-01
[W-14457826] Add RSS link for release notes

jdomingu:
This PR adds an RSS link on pages that include the `:page-subheading: rss` attribute. 

Screenshot 2024-02-01 at 12 10 35 PM

You can see the companion PR that adds that attribute to the release notes by-date page here:

https://github.com/mulesoft/docs-release-notes/pull/7863

Note that I've also added a tracking link modeled after our GitHub link, however I can't currently confirm the data is appearing until my access request goes through. Added a note to our existing tracking data spike to follow up on this.

### DIFF
--- a/src/js/05-analytics.js
+++ b/src/js/05-analytics.js
@@ -4,4 +4,6 @@
   if (!window.analytics) return
   const trackGithub = () => window.analytics.track('Clicked GitHub Link', { url: window.location.href })
   document.querySelectorAll('.github').forEach((githubLink) => githubLink.addEventListener('click', trackGithub))
+  const trackRSS = () => window.analytics.track('Clicked RSS Link')
+  document.querySelectorAll('.release-notes-rss').forEach((rssLink) => rssLink.addEventListener('click', trackRSS))
 })()
--- a/src/partials/subheading/do-rss.hbs
+++ b/src/partials/subheading/do-rss.hbs
@@ -1,0 +1,9 @@
+<a class="link-subheading align-center flex" href="https://docs.mulesoft.com/releases.rss" title="Add this URL to your RSS feed reader to stay up to date with MuleSoft releases.">
+    <svg width="24" height="24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 504.4 504.4" xml:space="preserve">
+        <path d="M377.6,0.2H126.4C56.8,0.2,0,57,0,126.6v251.6c0,69.2,56.8,126,126.4,126H378c69.6,0,126.4-56.8,126.4-126.4V126.6
+            C504,57,447.2,0.2,377.6,0.2z M136.8,409c-23.2,0-42-18.8-42-41.6c0-23.2,18.8-41.6,42-41.6c23.2,0,42,18.8,42,41.6
+            C178.8,390.2,160,409,136.8,409z M242,408.2c0-40-14.8-76-42.4-103.6c-28-28-63.6-42.4-103.6-42.4v-60.4
+            c112,0,206.4,94.4,206.4,206.4H242z M348.8,408.2c0-140-112.8-252.8-252.8-252.8V95c172,0,313.2,141.2,313.2,313.2H348.8z"/>
+    </svg>
+    Release Note RSS Feed
+</a>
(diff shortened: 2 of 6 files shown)


src/partials/subheading/do-rss.hbs
```diff
@@ -0,0 +1,9 @@
+<a class="release-notes-rss link-subheading align-center flex" href="https://docs.mulesoft.com/releases.rss" title="Add this URL to your RSS feed reader to stay up to date with MuleSoft releases.">
+    <svg width="24" height="24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 504.4 504.4" xml:space="preserve">
+        <path d="M377.6,0.2H126.4C56.8,0.2,0,57,0,126.6v251.6c0,69.2,56.8,126,126.4,126H378c69.6,0,126.4-56.8,126.4-126.4V126.6
+            C504,57,447.2,0.2,377.6,0.2z M136.8,409c-23.2,0-42-18.8-42-41.6c0-23.2,18.8-41.6,42-41.6c23.2,0,42,18.8,42,41.6
+            C178.8,390.2,160,409,136.8,409z M242,408.2c0-40-14.8-76-42.4-103.6c-28-28-63.6-42.4-103.6-42.4v-60.4
+            c112,0,206.4,94.4,206.4,206.4H242z M348.8,408.2c0-140-112.8-252.8-252.8-252.8V95c172,0,313.2,141.2,313.2,313.2H348.8z"/>
+    </svg>
+    Release Note RSS Feed
```
cheungaryk, line 8: should this be "Release Note**s** RSS Feed"? I know the title of that page is "Release Note Summary by Month", without the s for some reason.

src/partials/subheading/do-rss.hbs
```diff
@@ -0,0 +1,9 @@
+<a class="release-notes-rss link-subheading align-center flex" href="https://docs.mulesoft.com/releases.rss" title="Add this URL to your RSS feed reader to stay up to date with MuleSoft releases.">
```
cheungaryk, line 1: just a question...`align-center` doesn't seem to be doing anything here - I removed it, and the element still looks the same. Does it do anything?

(I am ok leaving it as-is though)

src/js/05-analytics.js
```diff
@@ -4,4 +4,6 @@
   if (!window.analytics) return
   const trackGithub = () => window.analytics.track('Clicked GitHub Link', { url: window.location.href })
   document.querySelectorAll('.github').forEach((githubLink) => githubLink.addEventListener('click', trackGithub))
+  const trackRSS = () => window.analytics.track('Clicked RSS Link')
+  document.querySelectorAll('.release-notes-rss').forEach((rssLink) => rssLink.addEventListener('click', trackRSS))
```
cheungaryk, line 8: can confirm that this gets sent to Segment, but only after I have removed the `not preview` condition seen here: https://github.com/mulesoft/docs-site-ui/blob/main/src/partials/head/head-scripts.hbs#L1C37-L1C50. Maybe we can improve this, like removing the `not preview` condition but label it instead, then in Segment, try to filter the labels to prevent them from being sent to the destinations.

cheungaryk (commented):
Looks good to me overall. I like the use of page attribute which allows users (aka writers) to opt in and opt out.

I left some questions in the comments.

src/js/05-analytics.js
jdomingu: Excellent. Thank you for checking. What I'm hearing is that the data will go through OK on the live site. I could go back and forth on including the segment script on the preview site. It's certainly convenient for testing, but I wouldn't want to pollute the data either.

I think I would want to add a `test` field or something similar to our events when they're submitted from the preview site. I'll punt on this for now, but add a note to the spike if that's OK with you.

src/partials/subheading/do-rss.hbs
jdomingu: Good catch. This was a copy-paste from another hbs template and I forgot to remove it. 

Removed!

src/js/05-analytics.js
cheungaryk: yes, a spike would be great. Don't want to block you here

cheungaryk (approved):

PR #655, opened 2024-02-02
[W-14954399] Swap deployment option PNGs with SVGs

jdomingu:
After rolling out the new lume icons, we got some feedback from writers about the deployment options tags listed on pages like this [one](https://docs.mulesoft.com/runtime-manager/deployment-strategies). These PNG images were missed in the search for SVG icons, so I'm updating those now.

Swapping some of the icons around as well based on a discussion with writers and design.

Before:
Screenshot 2024-02-02 at 3 19 27 PM

After:
Screenshot 2024-02-09 at 12 41 37 PM

Screenshot 2024-02-09 at 1 02 55 PM

Note that the cloud-ide and desktop-ide badges are not being touched per a discussion with design and writers.


cheungaryk (commented):
Implementation looks good, but in the local build, I see that cloudhub and cloudhub 2 icons are different. The following screenshot is what you put in the description:

Screenshot 2024-02-09 at 12 41 37 PM

In the local preview site (http://localhost:8080/elements/miscellaneous.html), I see the following set:

Screenshot 2024-02-09 at 2 30 35 PM

jdomingu:
Yikes, thank you so much for catching that. Looks like I didn't push my last commit. Would you mind trying now?

cheungaryk:
> Yikes, thank you so much for catching that. Looks like I didn't push my last commit. Would you mind trying now?

looks good now 😄

cheungaryk (approved):

PR #658, opened 2024-02-07
[W-14565448] Use updated heading sizing from design, remove two-tone coloring from…

jdomingu:
This PR switches to the new heading sizes from design. The main site does not seem to use H4 through H6, so unfortunately the sizes of those are very close. To make those easier to differentiate, those headings are in a different color.

Screenshot 2024-02-07 at 11 50 10 AM

Note that this also removes the two-tone coloring from the "MuleSoft Documentation" heading on our homepage, which was also requested by design.


cheungaryk (approved):
Validated locally via http://localhost:8080/elements/miscellaneous.html

PR #662, opened 2024-02-09
Docs homepage button should use Salesforce styles

jdomingu:
This PR updates the "New to MuleSoft" button on our homepage to use Salesforce styles. Currently, the button source is asciidoc in the docs-general repo [https://github.com/mulesoft/docs-general/blob/latest/modules/ROOT/pages/index.adoc](here).  To minimize duplicated CSS, I moved the button to a template in this repo directly, which lets us fully control the HTML structure and add classes more easily. (It looks like it is possible to use roles in the asciidoc to add css classes, but I wanted to avoid having to touch the JP source content directly as that seems like a bad practice--and it also seems harder to maintain going forward.)

**Testing**

Ran a full local build with these changes to ensure that the content displays correctly (regardless of whether you remove the `#cta` button from the asciidoc). 

Note that I _am_ removing the button from the English source file just to keep things tidy. PR here:
https://github.com/mulesoft/docs-general/pull/218/files


cheungaryk (approved):
looks good. Only thing is that in the JP preview site `npm run preview-jp`, the button is not localized, but a local build is localized. 🚢

jdomingu:
Oh good, it's not just me. I was seeing some odd behavior where the preview-jp site seemed to only partially replace text, but after an `rm -rf build public node_modules`, it seemed to work fine so I chalked it up to my environment. If you're seeing it too, then maybe it's worth looking at it more closely--I suspect there's something amiss with the live reload.

That said, merging this for now...

PR #668, opened 2024-02-23
[W-15032649] make single version selector into badge

cheungaryk:
ref: 
- W-15032649
- [W-12145035](https://gus.lightning.force.com/a07EE00001FNd4PYAT)

This PR turns version selector that has a single version into badge:

  Screenshot 2024-02-23 at 3 09 59 PM

I also increase the dropdown height to match the badge height as well.

Finally, I changed the nav version label from `<label>` to `<span>` to resolve an error reported by Chrome.


jdomingu:
Hey! Ran this locally and I'm seeing some odd behavior in the footer. The very bottom links no longer seem clickable for some reason and sometimes when I hover I see this:

Screenshot 2024-02-28 at 10 27 21 AM

Screenshot 2024-02-28 at 10 26 55 AM

Would you mind trying to reproduce this? Maybe it's just me?

Unrelated, although I like that you tried to differentiate these single-version badges from the drop-downs, I think it might be confusing for users since it matches the button style and we kind of reserve that shade of blue for clickable things. How would you feel about using one of the other colors from our styles? 

Screenshot 2024-02-28 at 10 32 57 AM
Screenshot 2024-02-28 at 10 34 45 AM

agansop (dismissed):

cheungaryk:
> Hey! Ran this locally and I'm seeing some odd behavior in the footer. The very bottom links no longer seem clickable for some reason and sometimes when I hover I see this:
> 
> Screenshot 2024-02-28 at 10 27 21 AM
> 
> Screenshot 2024-02-28 at 10 26 55 AM
> 
> Would you mind trying to reproduce this? Maybe it's just me?
> 
> Unrelated, although I like that you tried to differentiate these single-version badges from the drop-downs, I think it might be confusing for users since it matches the button style and we kind of reserve that shade of blue for clickable things. How would you feel about using one of these other colors? (The colors are cloud-blue-95, and aluminum-1 with aluminum-3 border to match the inline code styles if that helps.)
> 
> Screenshot 2024-02-28 at 10 32 57 AM Screenshot 2024-02-28 at 10 34 45 AM

@jdomingu 

I can't reproduce this for the footer links including the social media links. All these links are clickable and functional to me.

I have updated the look & feel:

Screenshot 2024-02-28 at 12 16 44 PM

src/js/01-nav.js
```diff
@@ -616,102 +616,114 @@
...
-          }
-        } else if (versionData === currentVersionData) {
-          navVersionMenu.appendChild(createElement('span.nav-version-label', 'Archived versions'))
+
+      let navVersionButton
+
+      if (versions.length === 1) {
+        navVersionButton = createElement('p.prev-flag')
+        navVersionButton.textContent = activeDisplayVersion
+        navVersionButton.title = `${activeDisplayVersion} is the current and only version of ${componentData.title}`
```
jdomingu, line 636: Sorry, missed this the first time around, but just noticed this is new text we'll need to localize.

No need to take care of it now necessarily, but would you mind adding it to the backlog? It might be easier to keep the message generic as well since interpolation doesn't work well with translations. Maybe just: "This is the current and only version."

jdomingu:
Thanks for doing that! Super bizarre that the bug is only happening in my environment. If it helps, here's what I see in the inspector:

Screenshot 2024-02-28 at 12 24 28 PM

That said, I'm happy to drop it if you're not seeing it. Maybe we can test again once it's up on the beta environment and see if anyone else can reproduce then?

jdomingu (approved):

src/js/01-nav.js
cheungaryk: sounds good. I will create a GUS ticket for localizing it and to make it more generic.

cheungaryk: GUS ticket created: [W-15145820](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001l7X6SYAU/view)

PR #672, opened 2024-02-27
[W-15133703] remove release notes pages from the JP site

cheungaryk:
ref: W-15133703

Add `isJPReleaseNotes(navItemData.content)` for the left nav item to check whether the left nav item is `リリースノート` ("release notes" in Japanese). If so, don't add it to the left nav. Doing so effectively hides the page so the users can't access it from the left nav.

To test, run `npm run preview-site-jp` and open [localhost:8080](). Click **Studio** and see that the release notes (aka リリースノート) is not listed.


jdomingu (approved):
Looks good! Confirmed the release notes page doesn't appear.

PR #673, opened 2024-02-27
[W-14075726] add hover link

cheungaryk:
ref: W-14075726

This PR adds the title property to any internal links, so that when a user hovers on one of those links, it will show the first paragraph as a tooltip.

Validation can be done on a beta site, for example the `custom-connectors` section: https://dev-docs-internal.kqa.msap.io/beta-hover-text/general/java-support#custom-connectors (all links except for "Deploy your applications" are testable). Ignore external links or dead links (for example, links to release notes since release notes is not part of this beta site). See the following screenshot:

Screenshot 2024-02-29 at 6 58 01 AM


jdomingu:
Nice work. Very nifty. No problems with the functionality itself. As I was looking at this, I had just two concerns (one real, one a nitpick).

Performance-wise, I would be concerned with rolling this out, as the title fetching is happening on load and requires spawning a new request for each link. Just for the shortish Java Support page, I'm seeing something like 25 new requests:

Screenshot 2024-02-29 at 10 34 26 AM

We'd have to test to be sure, but I think that for something like the incredibly long connector reference topics, this might be a dealbreaker. 

As for the nitpick, I think 512 characters seems very long for a tooltip. Holding your mouse in place long enough to read something like this for example isn't necessarily the best experience:

Screenshot 2024-02-29 at 10 15 28 AM

If we roll this out, I would recommend cutting down to something like 150 characters.

All to say, super sorry to leave this in limbo, but would you mind if we discussed this as a team before rolling it out?

cheungaryk:
> Nice work. Very nifty. No problems with the functionality itself. As I was looking at this, I had just two concerns (one real, one a nitpick).
> 
> Performance-wise, I would be concerned with rolling this out, as the title fetching is happening on load and requires spawning a new request for each link. Just for the shortish Java Support page, I'm seeing something like 25 new requests:
> 
> Screenshot 2024-02-29 at 10 34 26 AM
> 
> (We could probably improve the performance by maintaining a dictionary and deduplicating requests, but that might still not be enough for topics with lots of links like a connector reference topic.) I think ideally we would test this a bit more to be safe.
> 
> As for the nitpick, I think 512 characters seems very long for a tooltip. Holding your mouse in place long enough to read something like this for example isn't necessarily the best experience:
> 
> Screenshot 2024-02-29 at 10 15 28 AM
> 
> If we roll this out, I would recommend cutting down to something like 150 characters.
> 
> All to say, super sorry to leave this in limbo, but would you mind if we discussed this as a team before rolling it out?

Of course we can wait on deploying this! However, I will be on leave after today.

I adjusted it to 255 characters before looking at your review. It's an arbitrary number but I can do 150 characters.

jdomingu:
Awesome! Thanks for doing that and for being so flexible!

cheungaryk:
> Awesome! Thanks for doing that and for being so flexible! 🥇 Added this to our standup parking lot.

I made this into 150 characters as you suggested, and also made some changes to the code so that if the anchor is on the same page, just retrieve that section from the document object directly. This should greatly reduce the number of fetches (can test in the beta site later). (see [74031c5](https://github.com/mulesoft/docs-site-ui/pull/673/commits/74031c573b6519545baacee94646213202dcdbab))

Thank you for taking this over!

cheungaryk:
I added caching the DOM for pages: [794d8f9](https://github.com/mulesoft/docs-site-ui/pull/673/commits/794d8f9e58441ed6fd885f1aeae00022eafff1f3), so now there should only be 1 request per page. However, I wasn't able to test in time in the beta site, but feel free to validate.

jdomingu:
Following up: after expanding the beta test site and trying this on a wide variety of topics, I unfortunately can't recommend we roll this out. Although it works fine for the vast majority of topics, there are still a fair number of very long topics in our topics that this ends up inflicting a massive performance hit on--even with the excellent improvements you made in the second revision.

Here's the [most extreme example] taking just under a minute to fully load and launching 1600 odd requests (including CSS and JS and whatnot):
Screenshot 2024-03-20 at 9 26 55 AM

PR #675, opened 2024-02-29
[W-15023019] add dynamic trending topics to JP site

cheungaryk:
ref: W-15023019

Make JP site to use dynamically generated trending topics, powered by Coveo.

Before merging, make sure that the recommended content is accurate (inaccurate content would show content in descending alphabetical order, for example Zuora -> XML -> Xero -> Workday -> ...). Since Coveo User Analytics was recently added in #669, it may take awhile for the content recommendation to be tuned correctly.

If the content recommendation is still bad after awhile, then please consult Coveo to make sure the machine learning model and pipelines are configured correctly.


jdomingu (approved):
Excellent! Looks great 👍

PR #676, opened 2024-02-29
[W-15152389] Fix: Element focus obscured at bottom of page for some screen resolutions

jdomingu:
This PR fixes an edge cases with some screen resolutions which prevents you from clicking on elements on the very bottom of the page, and displaying a title tooltip:

Screenshot 2024-02-29 at 10 56 29 AM

Steps to reproduce (on another branch without this fix):

1. Open the browser developer tools. (Right-click anywhere in the page and click Inspect.)
1. Toggle the device toolbar (Ctrl/Cmd + Shift + M)
1. Set the resolution to 1200x1200.
1. Scroll to the bottom of the page and hover over any of the social media icons.

**Testing**

Confirmed at multiple resolutions that the bug no longer reproduces. Confirmed that the display of the version badges was not impacted by the change and you can still view their title attributes on hover. Also, tested in a full build to confirm this was fixed with a full left nav.


cheungaryk (approved):

cheungaryk:
validated locally based on instructions provided 🎆

PR #725, opened 2024-05-30
[W-15051462] fix minor bug for breadcrumbs + reorganize JP preview site

cheungaryk:
ref: W-15051462

- fix minor issue for breadcrumbs so if the breadcrumbs aren't present on the site (for example, in the homepage), the script doesn't need to execute further
- move the JP preview site files into the correct location so the preview site can find them
- update header


jdomingu (approved):
Looks good! Confirmed this locally. Merge away!

Probably unrelated: Do you see this in the trending topics recs on your end?

Screenshot 2024-05-30 at 10 45 00 AM

cheungaryk:
> Looks good! Confirmed this locally. Merge away!
> 
> Probably unrelated: Do you see this in the trending topics recs on your end?
> 
> Screenshot 2024-05-30 at 10 45 00 AM

Great! Thanks for the approval. The issue with the "no title" error is due to the incident with the JP redirect. Once that issue is resolved, this issue will be resolved as well.

PR #727, opened 2024-06-03
[W-15921470] trending topics fallback

cheungaryk:
ref: W-15921470

- add fallback links for when trending topics (from Coveo) aren't returning any results (or bad results)
- update JP site marketing header

### How to test

1. Run `npm run preview-jp`
2. Open http://localhost:8080/jp on your web browser (note: not localhost:8080. I will have to fix that later). You should see initially that the trending topics box shows a bunch of "no titles", and then after a few seconds, they changed into a bunch of real links. These links are hardcoded in a JSON file in case these results are either "no titles" or if there are no results at all


src/js/25-trending-topics-fallback.js
```diff
@@ -0,0 +1,47 @@
...
+    const atomicRecsResult = recsListShadowRoot.querySelector('atomic-recs-result')
+    if (atomicRecsResult) {
+      const recsResultShadowRoot = atomicRecsResult.shadowRoot
+      const atomicText = recsResultShadowRoot.querySelector('atomic-text')
+      return atomicText?.value !== 'no-title'
+    }
+    return true
+  }
+
+  setTimeout(() => {
```
jdomingu, line 33: I notice that the coveo examples use something like `await customElements.whenDefined('atomic-search-interface')` to listen for the coveo elements loading. I wonder if that would work? If so, I think it would really improve the experience, as a four second content swap feels quite long.

src/locales/trending-topics-fallback.json
```diff
@@ -0,0 +1,68 @@
+{
```
jdomingu, line 1: You might consider moving the fallback links out of json and into plain html in the trending topics hbs file, then toggling between the fallback div and the atomic div with css. 

Since we're only using the trending topics on the one page, it would be nice to pull it out of the js bundle. At the end of the day, it's a small amount of content though, so not a huge deal.

jdomingu (dismissed):
This all works for me locally 👍 

Added some suggestions below, but they're not dealbreakers so feel free to merge if they don't seem worth it to you.

src/locales/trending-topics-fallback.json
cheungaryk: You are totally right. I haven't worked with the UI bundle for so long, I forgot I can do that. Let me change them to HTML

src/js/25-trending-topics-fallback.js
cheungaryk: let me try this

src/locales/trending-topics-fallback.json
cheungaryk: I moved these links into .hbs. Also made them into relative links so they don't work in the preview site, but should work in a regular site.

Screenshot 2024-06-04 at 4 35 34 PM

src/js/25-trending-topics-fallback.js
cheungaryk: So technically, this works (as in it doesn't generate an error in the console), but it doesn't do what is intended in this case. That's because atomic-search-interface is always defined quickly when the page loads, but it takes a little bit of time to generate the shadowRoot content which is what I want to check.

I set it to 4 seconds initially in case the internet is slow, but I think 1 second may be ok.

jdomingu (approved):
Awesome work! Thanks for trying my suggestions.

PR #736, opened 2024-06-14
[W-15230132] Fix: Archive connectors landing page missing connectors

jdomingu:
The [connectors landing page](https://archive.docs.mulesoft.com/connectors/) for the archive site does not show all the connectors available. This was a strange one to debug as side-by-side comparisons of the content did not yield any noticeable differences, and the code in this repo does not have any special template for the archive connectors table. 
It turns out that this issue is caused by the Antora Atlas extension. The connectors list is generated based on the latest version of components (this list is created by Antora at build-time), and this list also includes versions of components from the [prod site's manifest](https://docs.mulesoft.com/site-manifest.json). These imported versions are always set as the latest version, which is a problem because imported versions do not include the connector attributes we were keying off to generate the connectors list.

TLDR: Filtering out imported connector versions and setting the latest connector version to a local version fixed the issue. 

**Testing**

Tested to confirm that full builds of the prod and archive site display all versions.

### DIFF
--- a/src/helpers/connectors.js
+++ b/src/helpers/connectors.js
@@ -5,7 +5,32 @@ module.exports = function ({
     root: { site },
   },
 }) {
-  return Object.values(site.components).filter(
-    ({ asciidoc = { attributes: {} } }) => asciidoc.attributes['page-connector-type']
-  )
+  function getLocalConnectorComponents (components) {
+    const connectors = {}
+
+    for (const component in components) {
+      if (component.endsWith('-connector')) {
+        const connector = components[component]
+        // Filter out connector versions that are missing the page-connector-type attribute.
+        // These are imported from the site manifest with the @antora/atlas-extension in the playbook,
+        // and should not display in the build.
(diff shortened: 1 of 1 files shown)


cheungaryk (approved):
I approved since all the connectors are present, but the resulting list has 2 veyr minor issues:
1. Some logo icons are broken. Probably not going to be fixed (aka can ignore) because the fixes are likely in the docs repos to correctly construct the image URLs, and writers are to leave the archived versions as-is
2. Some connectors appear out of order. For example, Marketo is at the top of the list even though it starts with an M, and CassandraDB is between AMQP and Apache Kafka. This is probably because these connectors have been renamed in the prod site to "Adobe Marketo" and "Apache Cassandra" respectively

OK to merge as it is though.

jdomingu:
Thanks for the thorough testing. The first one I noticed and agreed, didn't see a good way to fix as the URLs resolve with a 200 status.

The second issue I should be able to fix without too much trouble. I'll see if I can add that to the followup PR for the popular connectors.

PR #739, opened 2024-06-15
W-16007932  Update regex to match two or three digit numbers between "." in versions

marinasasso:
[W-16007932] This change will make the version regex in cleanTitle() and isVersion() match versions up to 999 in the second and third places.
Currently having a two digit version in the last place causes the regex to not work as expected, for example:
API Manager version: 2.4.30


cheungaryk:
Thanks Marina. I used your PR as the base and submitted and merged #741 (sorry I originally thought you closed this PR, so I started a new one)

PR #742, opened 2024-06-19
@W-16025225 - Fix assorted connectors landing page bugs

jdomingu:
This PR includes the following changes:

* Fixes an issue where popular connectors for JP do not display images. (This and the next item were simply applying the revised Antora helper from #736 and doing due diligence to test across sites.)
* Fixes an issue where popular connector links for JP point to the English site. (Compare hover URLs in the images below. After the fix, the URLs show proper relative paths. Note that on archive builds, these properly link to specific version URLs, not latest.)
* Enables the popular connectors section for the archive site.
* Fixes an issue where the list of connectors is not alpha-sorted. For example, the archive site shows Marketo first (because the component name for newer versions is Adobe Marketo.)
* Updates the `caniuse` dependency after being prompted to run the update browserlist command. 
* Adds an option to override connector icons. For the archive site, several of the icons on the landing page are broken. However, you cannot change the group and asset ids to point to the new paths as this breaks the exchange URLs. 
  * Adding `page-connector-icon-override: true` for a connector will default to using the generic icon instead of the would-be broken default icon path.
  * Adding `page-connector-icon-url` allows you to provide a hardcoded URL.
  * You can also see these attributes in the `ui-model.yml` file for the preview site.
  * After this change goes out, I will merge https://github.com/mulesoft/docs-connectors/pull/2610. (See before and after images in this PR.)

Before and after of popular connectors issues:

Screenshot 2024-06-18 at 3 57 17 PM
Screenshot 2024-06-18 at 4 17 18 PM

Before and after of sorting issue:

Screenshot 2024-06-18 at 6 59 56 PM
Screenshot 2024-06-18 at 7 00 12 PM

**Testing**

Tested with local playbook builds of main, jp, and archive.


agansop (dismissed):

cheungaryk (approved):
validated via local builds for prod, archive, and jp. All look good! Thank you

PR #777, opened 2024-08-27
[W-15177355] Update collapsible block styling

jdomingu:
Match Salesforce styling for accordion items, or as Asciidoc calls them, [collapsible blocks](https://docs.asciidoctor.org/asciidoc/latest/blocks/collapsible/).

**Testing**

* Tested in Chrome and Safari.
* Tested across multiple screen resolutions.
* Added example collapsible blocks to the preview site for testing.

Including a few screenshots of the output below.

Stacked blocks:

Screenshot 2024-08-27 at 3 15 46 PM

Hover styling:

Screenshot 2024-08-27 at 3 17 47 PM

Wrapped title and heading in block:
Screenshot 2024-08-27 at 3 18 02 PM

### DIFF
--- a/preview-site-src/elements/miscellaneous.adoc
+++ b/preview-site-src/elements/miscellaneous.adoc
@@ -58,10 +58,49 @@ Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor i
 
 Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
 
-== Details
+== Collapsible Blocks
 
-.Click to expand the details element
+.Stacked Collapsible Block 1 Title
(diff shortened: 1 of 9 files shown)


cheungaryk (dismissed):
Congrats on the PR number 777! 🎆 

I have tested the changes successfully. I have approved it.

Not a blocker, but the table inside this collapsible block, in a small browser screen size, doesn't have the headers on the left:

Screenshot 2024-08-27 at 3 59 17 PM

This is how a table with header should look like (from http://localhost:8080/elements/table.html) for reference:

Screenshot 2024-08-27 at 3 58 19 PM

jdomingu:
Thanks for the thorough testing as always! I though it looked a bit funny during testing as well, but left it since it matched the current behavior [here](https://docs.mulesoft.com/api-manager/latest/proxy-latest-concept).

I went ahead and tracked down the issue--our code wasn't finding any `innerText` for the headers, which makes sense since the tables aren't visible on page load. Switching to `textContent` fixes the issue. While I was at it, I reduced the table margin so it wouldn't look so wonky:

Screenshot 2024-08-27 at 5 27 58 PM

As a side-note, I don't know why we chose not to display the mobile table toggle for nested elements, but it was done intentionally so I left it. You can see where that's set [here](https://github.com/mulesoft/docs-site-ui/blob/W-15177355-update-collapsible-block-styling/src/js/20-table.js#L128).

cheungaryk (approved):
nice! 🚢 

> As a side-note, I don't know why we chose not to display the mobile table toggle for nested elements, but it was done intentionally so I left it.

I don't remember why it was done that way, it's been awhile. It's fine to change it...

PR #785, opened 2024-09-10
[W-16646670] coveo migration

cheungaryk:
ref: W-16646670

Moving Coveo search from the mulesoft org to the salesforce org (dev). Most of the work is renaming fields and changing the org. 

Coveo User Analytics (UA) is not in scope, so there are no changes in the UA scripts until after the moratorium ends.

### How to test

In 1password, search for **coveo token**. Copy the dev secrets for EN and JP site and export them as these variables (I have these added to my .zshrc file for convenience):

```sh
export COVEO_API_KEY=xx13....
export COVEO_API_KEY_JP=xx8...
```

Then run `npm run preview` and perform a search (example: http://localhost:8080/general/search.html#f-mulesoftproduct=Connector%20DevKit). Note that the local preview site doesn't have all the products and versions so most of the version facets (the ones that appear when you select a product) don't work. One product that works is **Connector DevKit** if you want to validate the version facets:

Screenshot 2024-09-10 at 7 23 06 AM

You can also run `npm run preview-jp` then go to http://localhost:8080/jp/general/search.html#f-mulesoftproduct=Connector%20DevKit to check on the JP results.

### DIFF
--- a/preview-site-src-jp/ui-model.yml
+++ b/preview-site-src-jp/ui-model.yml
@@ -27,6 +27,8 @@ site:
           - content: API ビルドの最初から最後まで
             url: /jp/general/api-led-overview.html
             items:
+              - content: ドキュメントの検索
+                url: /jp/general/search.html
               - content: ステップ 1. 前提条件
                 url: /jp/general/api-led-prerequisites.html
               - content: ステップ 2. API 仕様の設計
--- a/src/js/21-coveo-facets.js
+++ b/src/js/21-coveo-facets.js
@@ -22,7 +22,7 @@
       if (formattedVersions) {
         const versionFacet = document.createElement('atomic-facet')
         versionFacet.setAttribute('allowed-values', formattedVersions)
-        versionFacet.setAttribute('depends-on-product', componentDisplayName)
+        versionFacet.setAttribute('depends-on-mulesoftproduct', componentDisplayName)
         versionFacet.setAttribute('facet-id', `${componentID}-version`)
         versionFacet.setAttribute('field', 'mulesoftversionwithlatest')
         versionFacet.setAttribute('heading-level', 2)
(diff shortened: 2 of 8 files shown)


jdomingu (approved):
Nice. Tested locally and everything seems to work.

A couple of questions for when we roll this out:
* When would we want to roll out the API key update? (I'm assuming we'd need to coordinate these two changes.)
* When will we want to switch from the Salesforce dev org to the prod org? (Or could we potentially switch to that directly since it looks like we have access?)

cheungaryk:
> When would we want to roll out the API key update? (I'm assuming we'd need to coordinate these two changes.)

I will swap up the API key by running this Jenkins job: https://jenkins.build.msap.io/job/engineering-services/job/Tooling/job/jenkins-credentials-automation/job/master/build?delay=0sec. Unfortunately, I don't think we can do different Coveo keys for different environments (like dev API key for lower envs and prod API key for higher envs), so I am just going to use the same key for all envs.

The timeline is:
1. I merge this PR and merge the playbook PRs
2. I run the Jenkins job
3. I wait for the next deployment to happen and validate it in the lower envs. Due to the moratorium, the prod sites are going to be unaffected, even on the Coveo side because I am using a different Coveo org while the prod sites will continue using the old org (mulesoft prod)

> When will we want to switch from the Salesforce dev org to the prod org? (Or could we potentially switch to that directly since it looks like we have access?)

We can make the switch once dev org is validated. I need to ask the Coveo people if they can promote my changes to the prod org so I don't have to recreate everything.

cheungaryk:
I will merge this after I get Marie Y from Salesforce to validate that my changes did not impact DSC/H&T search of the MuleSoft pages.

jdomingu:
Awesome. Thanks for helping me understand the process here. That sounds good!

PR #806, opened 2024-10-09
[W-16947933] archive site toolbar enhancements

cheungaryk:
ref: W-16947933, https://github.com/mulesoft/docs-general/pull/274

please also review and approve https://github.com/mulesoft/docs-general/pull/274.

Add some logic to make archive site landing page compatible with home.hbs which removes the empty toolbar near the top of the homepage. The toolbar is still present in the mobile screen size of the homepage, and it is present in non-hompages (any screen size).

Also added an extra `<ol>` element to the mobile screen of the archive site homepage so that the **menu** button aligns to the left of the page rather than in the center.

Screenshot 2024-10-09 at 4 40 45 PM

Screenshot 2024-10-09 at 4 41 03 PM

Validated by exporting the build UI zip file and then build EN and archive sites with it.

release.js has no changes, it was just formatted.

### DIFF
--- a/gulp.d/tasks/release.js
+++ b/gulp.d/tasks/release.js
@@ -63,7 +63,15 @@ const createPR = async ({ octokit, owner, repo, tagName, ref, sites, secretKey, 
   const newBranchName = tagName
   await createNewBranch({ octokit, owner, repo, ref, newBranchName })
   await updateContent({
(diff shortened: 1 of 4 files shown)


jdomingu (approved):
Excellent. That looks much better 👍

PR #819, opened 2024-10-31
[W-15153471] smart snippet

cheungaryk:
ref: W-15153471

add smart snippet to site search results:

Screenshot 2024-10-31 at 1 43 27 PM

note that if you want to test, you have to switch to salesforce-prod org ID and tokens (Set these in ~/.zshrc, see Coveo tokens note in 1password) because the smart snippet model is not present in salesforce-dev as I don't have permission to create 1 there (and also we are reaching the model limit with the license in that Coveo org)


cheungaryk:
We are no longer adding smart snippet. Closing

PR #832, opened 2024-12-20
[W-17043241] Add support for pills, starting with two Anypoint pricing pills

jdomingu:
This PR adds support for pills that display below the topic H1. This is very similar to the existing [deployment options](https://docs.mulesoft.com/runtime-manager/deployment-strategies) pills, but made more generic to support any type of pill that writers want to include in the future. 

(In other words, we have to keep the `page-deployment-options` attributes as they're listed in a lot of topics, but adding new pills to deployment options when the text and icons have nothing to do with deployments seems a bit odd. Hence breaking this out separately.)

As a nice byproduct, making this more generic means that we can push text to the messages file for future translations and don't have to create a new hbs file for each new page attribute.

Here's what these pills look like:

Screenshot 2024-12-19 at 5 47 20 PM

Caveats:
* We don't have any JP strings for this yet, and the handoff just happened so we won't get a chance to submit strings for awhile. Pricing tiers for Anypoint are going to be simplified soon though, so the two pills will collapse to one within the next few months. Meaning, we might not want to hand off anyway as the text might change. Will follow up with Duke on this one.
* Because I wanted the pills to be as generally applicable as possible, I've tested support for multiple link types:
    * Relative links and absolute links. Work normally. We don't touch these URLs.
    * Links to the root (i.e. that start with `/`). For cases where we can't guarantee that all pages will have the same relative path to the linked topic--especially when linking from multiple guides--we can link to the root. I added a little code to confirm this works with JP builds as well by prepending `/jp` to links. The downside of using these links is that they will not work in local playbook builds.

Testing

Confirmed this displays the same as the deployment options pills.
Confirmed you can add multiple attributes or one attribute with the same result. (Ordering does not matter.)
Confirmed the functionality on the JP preview site (though we don't have any JP strings).
Tested in local build. (Links will not work as mentioned above.)


preview-site-src/elements/admonition.adoc
```diff
@@ -2,6 +2,7 @@
 :keywords: admonition, admonitions, warning, section, notes
 :page-component-name: elements
 :page-support-status: eolScheduledFeature
+:page-info-pills: anypoint-pricing-core, anypoint-pricing-titanium, undefined-pill
```
cheungaryk, line 5: is this here to demonstrate that if the pill is undefined, it's not added to the page?

cheungaryk (approved):
LGTM! 1 minor comment

preview-site-src/elements/admonition.adoc
jdomingu: Yep! Exactly.

PR #851, opened 2025-02-19
[W-17466952] Add learning map support

jdomingu:
Add support for Salesforce-style learning maps. Here's an [example from the Salesforce help](https://help.salesforce.com/s/articleView?id=ind.mfg_get_started_service.htm&type=5).

Note that I ended up not creating a separate layout for this as it would have duplicated the default layout exactly. Instead, I'm just adding a class to the article element based on the `page-article-style` attribute and keying off that for styling.

Here's what that looks like:

Screenshot 2025-02-19 at 10 06 31 AM

**Testing**

Tested in preview builds.
Tested in local English build. (Will work with writers to add to docs-general later.)


cheungaryk (approved):
looks good! However, `:page-article-style: learning-map` didn't work for my local build, not sure why. I did validate the UI by manually adding learning-map as a class.

PR #855, opened 2025-02-20
[W-17627075] Fix: Nav version should not contain nested interactive elements

jdomingu:
Fixes this accessibility issue: https://knowledge.evinced.com/system-validations/nested-interactive

This is caused by the interactive button containing an interactive tooltip dot. From what I can tell, the offending button here is purely decorative: we remove it from tab navigation with `tabindex=-1`, and drop down part of the version control is outside of it. 

Switching from a button to a div seems to fix the problem. Also, though it wasn't strictly necessary, I went ahead and renamed the class to reflect the change. I.e.:

`button.nav-version-button` --> `div.nav-version-wrapper`

**Testing**

Tested in local preview.
Confirmed version drop-down still works.
Confirmed keyboard navigation remains the same.

### DIFF
--- a/src/css/specific/nav.css
+++ b/src/css/specific/nav.css
@@ -309,7 +309,7 @@ svg.nav-icon {
   }
 }
 
-.nav-version-button {
+.nav-version-wrapper {
   align-items: center;
   background: 0 0;
   border: 1px solid var(--aluminum-4);
--- a/src/js/01-nav.js
+++ b/src/js/01-nav.js
@@ -14,7 +14,7 @@
   const addCurrentVersionIndicator = (parentElement, className) => {
     if (!isArchiveSite()) {
       if (!isToolTipDot(parentElement.firstChild)) {
-        const tabIndex = parentElement.classList.contains('nav-version-button') ? 0 : -1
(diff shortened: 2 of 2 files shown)


cheungaryk (approved):
confirming that in local preview build, the TOC keyboard navigation works as usual. However, when I focus on the pillbox using keyboard (tab key), I see the focused state looking like this:

Screenshot 2025-02-20 at 3 57 18 PM

jdomingu:
I _think_ that's how it is now, but do let me know if what you're seeing is different than on prod. Here's what I see:
Screenshot 2025-02-20 at 4 09 34 PM

If there's something else that makes more sense, happy to change it though.

cheungaryk:
> I _think_ that's how it is now, but do let me know if what you're seeing is different than on prod. Here's what I see: Screenshot 2025-02-20 at 4 09 34 PM
> 
> If there's something else that makes more sense, happy to change it too.

That's ok to leave it like that then. Probably something too minor to change anyway unless it's really easy

jdomingu:
Sounds good. Thanks for testing!

PR #859, opened 2025-02-26
[W 17900983] add agentforce promo banner

cheungaryk:
ref: [W-17915813](https://gus.lightning.force.com/a07EE00002AMgbmYAD)

Add agentforce promo banner to the top of the docs site:

Screenshot 2025-02-26 at 7 38 38 AM

To make this change, I added the `promotion-banner=true` to the marketing header API endpoint, and made adjustments to any CSS (toolbars and banners in particular) that has to do with nav heights.

### how to test

easiest way to test is using local build.

1. run `rpm run preview`
2. open an incognito window (this is because with incognito, it's easier to reset the promo banner behavior) and open http://localhost:8080/
3. resize the windows and scroll up and down for each different screen sizes. The header elements should look "normal"
4. close the promo banner by clicking the **X** on the upper right corner. Then repeat step 3
5. close all incognito windows, open another incognito window and go to http://localhost:8080/elements/code-block.html. This has more stuff like toolbars and notice banners
6. repeat steps 3 and 4


jdomingu (approved):
Awesome work! Tested locally at multiple screen resolutions and it all worked great. I also confirmed that the banner stays hidden after dismissing it and reloading the page.

Ship it! 🚀

PR #870, opened 2025-04-02
[W-17627001] Fix: nav-version-menu includes listbox aria role without required children

jdomingu:
Here is an example of the error from Evinced: https://app.evinced.com/properties/cb70a00e-5e97-494b-ae37-ae213b71d31c/scans/a6aeafeb-512f-457d-b00a-bfae0162063d/issues/e221cdff-a809-4abb-b139-8fb4af6c5df8

`Element has children which are not allowed: span[tabindex], button[tabindex]`

Here is their reference page: https://knowledge.evinced.com/system-validations/aria-required-children

In other words, elements with the `listbox` role can't have arbitrary children--they need to be children with the `option` role per this [page on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role).

I've updated the structure of our menus to match the [Listbox example with grouped children](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) on the W3C Aria site.

To make it a bit easier to see the change in HTML structure, here's a diff of the rendered HTML:
Screenshot 2025-04-01 at 7 13 43 PM

**Testing**

Tested with a full build of the English and Archive sites.

### DIFF
--- a/src/css/components/tooltip.css
+++ b/src/css/components/tooltip.css
@@ -48,7 +48,7 @@
   }
 
   &.tooltip-dot-nav-version {
-    margin-left: 6px;
+    margin-left: -12px;
     margin-top: 8px;
   }
 }
--- a/src/css/specific/nav.css
+++ b/src/css/specific/nav.css
@@ -391,13 +391,14 @@ svg.nav-icon {
       max-height 0s var(--transition-speed-sm);
   }
 
-  & button,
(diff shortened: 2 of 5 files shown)


cheungaryk (approved):
LGTM!

PR #951, opened 2026-08-19
fix(security): patch critical Zip Slip vulnerability in transitive deps

Non3e:
## Summary

Resolves the **critical** [CVE-2020-12265](https://github.com/advisories/GHSA-x2mc-8fgj-3wmr) (Zip Slip / Arbitrary File Write) vulnerability in the `decompress` transitive dependency.

### Changes
- Added `overrides` in `package.json` to force `decompress >= 4.2.2` and `http-cache-semantics >= 4.1.1`
- Regenerated `package-lock.json` with patched transitive dependency versions

### Why overrides instead of direct upgrade?
`gulp-imagemin` 8+ migrated to ESM-only, which would require rewriting all `require()` calls in `gulp.d/`. The `overrides` approach patches the transitive vulnerability without any breaking changes to the build system.

### Before/After
| Metric | Before | After |
|--------|--------|-------|
| Critical vulns | 1 | 0 |
| High vulns | 28 | ~28 (transitive reporting changes) |

### Testing
- `npm install --package-lock-only` succeeds
- No changes to build scripts or runtime behavior
- Only `package.json` (4 lines) and `package-lock.json` (dependency resolution) modified

Related: #840 (security hygiene)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
       "name": "mulesoft-docs-site-ui",
       "devDependencies": {
         "@asciidoctor/core": "~2.2",
-        "@asciidoctor/tabs": "*",
+        "@asciidoctor/tabs": "latest",
         "@docsearch/css": "^4.0.0-beta.7",
         "@docsearch/js": "^4.0.0-beta.7",
         "@mulesoft/lume-icon": "file:vendor/mulesoft-lume-icon-0.1.47.tgz",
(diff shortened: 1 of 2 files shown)


Non3e:
I have read the CLA Document and I hereby sign the CLA

jdomingu (approved):

jdomingu (changes requested):
Actually, given that this isn't the usual renovate bot, I think we need to look at this more closely before merging. Blocking for now.

cheungaryk:
Hi @Non3e, thanks for your contribution but since we have upgraded in gulp-imagemin in #940 which no longer depends on `bin-build` or `decompress`, this PR is obsolete and we will close it.

